PR #11, opened 2025-06-13
Feature/frontend app scaffold

XanderVertegaal:
Adds an empty Nuxt application to `/frontend`. This PR also contains two Dockerfiles that allow you to run both a dev server (with live reload) and an optimized, static prod build in containers. Check the README for installation instructions.

I've also added a linter (ESLint) and enabled Prettier (by adding an empty `.prettierrc`).

Environment variables for the containers are not handled properly yet, but I'll leave that for a future issue.


frontend/app.vue
```diff
@@ -0,0 +1,6 @@
+<template>
+  <div>
```
XanderVertegaal, line 2: Edit this file (e.g. add a `<p>Hello, World!</p>`) to see whether live-reloading works.

frontend/eslint.config.mjs
XanderVertegaal, line 1: These are linting options. I've simply added the config we have for DIAPP here (with a few stricter options). If we don't like certain rules, we can configure them here.

frontend/nuxt.config.ts
```diff
@@ -0,0 +1,56 @@
...
+export default defineNuxtConfig({
+  compatibilityDate: "2025-05-15",
+  devtools: { enabled: true },
+  modules: ["@nuxt/eslint", "@nuxt/test-utils"],
+
+  // Disable server-side rendering.
+  ssr: false,
+
+  runtimeConfig: {
+    public: {
```
XanderVertegaal, line 11: The DIAPP has these env variables. They are automatically taken from an `.env` file at build time, so I only have to initialise them here. I'll save adding an `.env` for a future PR.

compose.yaml
```diff
@@ -0,0 +1,47 @@
...
+    container_name: mr-nginx
+    restart: unless-stopped
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf:ro
+    ports:
+      - "5000:80"
+
+  vue-dev:
+    image: mr-vue-dev
+    container_name: mr-vue
```
XanderVertegaal: This naming scheme for the services and the images is something I have started to use for RSLab projects. I'm not married to this, so if anyone prefers something else, I'm happy to change it :)

Whatever the naming scheme we decide on, I would strongly suggest that we at least pick identical `container_name`s for prod and dev containers. That way, other containers can refer to a service without having to know whether that container is running in prod or dev mode.

frontend/Dockerfile.prod
```diff
@@ -0,0 +1,28 @@
...
+
+# Copy built application files from the builder stage into the NGINX server that will server them.
+COPY --from=builder /app/.output/public /usr/share/nginx/html
+
+# Copy the NGINX configuration file into the container.
+COPY ./nginx.conf /etc/nginx/conf.d/default.conf
+
+# Expose the port for the NGINX server.
+# Does not actually expose the port but documents it.
+EXPOSE 3000
```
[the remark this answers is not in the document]
XanderVertegaal: The global NGINX server for the entire compose network (`/nginx.conf`) listens on port 80, but the NGINX server that handles the production frontend server (`/frontend/nginx.conf`) listens on port 3000, so `EXPOSE 3000` is correct.

README.md
```diff
@@ -1,5 +1,80 @@
...
+
+The differences are outlined below.
+
+| | Local | Docker dev | Docker prod |
+| --- | --- | --- | --- |
+| PostgreSQL database | On host machine | In container | In container |
+| Frontend server | Nuxt development server | Nuxt development server in container | NGINX server serving a minimized static build |
+| Backend server | TBD | TBD | TBD |
+| Live-reloading on code changes | ✔ | ✔ | ❌ |
+| Ports available | 3000 (frontend) | 8000 (NGINX), 3000 (frontend) | 5000 (NGINX) |
```
tijmenbaarda: As far as I can see, for the dev docker compose, it is also port 5000 that is exposed for nginx, not 8000.

```suggestion
| Ports available | 3000 (frontend) | 5000 (NGINX), 3000 (frontend) | 5000 (NGINX) |
```

tijmenbaarda (approved):
Looks good to me, apart from one small thing in the README!

I think it would be good if @miggol could take a last look at the Docker setup before this is merged.

compose.yaml
```diff
@@ -0,0 +1,47 @@
...
+    restart: no
+    profiles: ["dev"]
+    build:
+      context: ./frontend
+      dockerfile: Dockerfile.dev
+    command: npm run dev
+    expose:
+      - "3000"
```
miggol, line 23: Do we need to expose 3000 if we can access it through nginx on port 5000 anyway?

compose.yaml
miggol: Reusing the same container name gives me this problem:

```
Error response from daemon: Conflict. The container name "/mr-vue" is already in use by container "a6a3ed590c0a2d4e1d2dbb7ca751f2275b383264c1ee901a556d4e0c239726a7". You have to remove (or rename) that container to be able to reuse that name.
```

How about we rename this one to `container_name: mr-vue-dev` and set `hostname: mr-vue` on both profiles so other services can find them by the same hostname?

frontend/app.vue
miggol: Nice! It works

nginx.conf
miggol, line 1: TODO: Configure logging

frontend/Dockerfile.prod
```diff
@@ -0,0 +1,28 @@
...
+# Install project dependencies.
+RUN npm install
+
+ENV NUXT_PUBLIC_BUILD_DATE="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+
+# Build the application in production mode.
+RUN npm run generate
+
+# Create a new stage for the production image.
+FROM nginx:alpine
```
miggol, line 21: Funky... Double nginx. Makes me wonder if this could this be shared by one instance?

Not something to investigate in this PR, though. Works fine like this.

miggol (approved):
Great start! Sorry for the late review.

My only hard issue is with `container_name`, approving otherwise

nginx.conf
miggol: Not for this PR (see #27)

compose.yaml
miggol: I do like the naming scheme also. I've started calling the container "Mister vue" in my head and I can't shake it

README.md
XanderVertegaal: Correct! Typo on my end.

compose.yaml
XanderVertegaal: No, you are definitely right that we don't need to expose 3000 here, but having it here provides us with direct access to the dev container without having to rely on the reverse proxy, so we can test it in isolation.

compose.yaml
XanderVertegaal: Oh nice, I didn't know about `hostname`. I'll add one to the NGINX container as well.

PR #26, opened 2025-06-26
feat: add custom user

### LINKED ISSUE #10: Custom Django User and UserGroup models

EdoStorm96:
fixes #10 

Again ... Just did the bare minimum.

I realize the issue says to also add a custom UserGroup model, but I am not sure how necessary this is? I haven´t needed this before ... Would I just extend the Django builtin Group model or create a new model?

Anyway, if you can explain to me the purpose of a custom group model, I am of course happy to add this!


backend/main/admin.py
```diff
@@ -0,0 +1,31 @@
+from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
+from django.db import models
+from django.forms.widgets import TextInput
+
+from .models import User
+
+
```
tijmenbaarda, line 8: Too many blank lines. Perhaps run `black`.

tijmenbaarda:
Looks good!

I don't think there is need for a custom group model at this point. From what I see it is not straightforward to customize the `Group` model [1] (unlike the `User` model).

I do think we will need to have a more elaborate group functionality in the future though, because the Django groups are very basic. In DIAPP the Group model was not customized, but instead a new group-like model was built from scratch (`Role`), with an intermediate model `RoleAssignment` to make it possible to assign a role to a user specifically for certain faculties or departments. Something like that could also be useful for MyResearch, but it is not necessary to do that now.

[1] https://forum.djangoproject.com/t/custom-group-model/30070/2

tijmenbaarda (approved):

backend/myresearch/settings.py
```diff
@@ -31,6 +31,7 @@
 # Application definition
 
 INSTALLED_APPS = [
+    "main",
```
XanderVertegaal: Nitpick: put `main` (and other apps added by us) below the built-in ones.

(Not a rule that is set in stone, but a custom I've seen over the years.)

PR #31, opened 2025-07-04
Add uu-bootstrap

tijmenbaarda:
This PR adds a basic layout using `cdh-vuelib` and `uu-bootstrap`. The header, navigation bar and footer were adapted from the example of DIAPP. I also included other static files, such as fonts and logos, from the DIAPP repository. No internationalization is added yet, so the strings are now just plain English (but there are only a few).


frontend/plugins/fontawesome.ts
```diff
@@ -0,0 +1,83 @@
+import { library } from "@fortawesome/fontawesome-svg-core";
+// eslint-disable-next-line import/named
+import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome";
+
+import {
+    faUser,
```
XanderVertegaal: We don't need all of these icons yet, so perhaps it's best to leave the ones we aren't currently using out.

frontend/plugins/fontawesome.ts
```diff
@@ -0,0 +1,83 @@
+import { library } from "@fortawesome/fontawesome-svg-core";
+// eslint-disable-next-line import/named
```
XanderVertegaal: Are you getting any warnings without this line? ESLint doesn't complain for me.

frontend/plugins/fontawesome.ts
```diff
@@ -0,0 +1,83 @@
...
+    faCircleXmark,
+    faGraduationCap,
+    faArrowRight,
+    faSkull,
+    faCrown
+);
+
+export default defineNuxtPlugin((nuxtApp) => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore Somehow it doesn't get that a Component is a Component
```
XanderVertegaal: TS is fine with this, so perhaps we can remove this @ts-ignore comment. (It was probably needed in the past, but whatever it was, it seems to be fixed now.)

frontend/components/layout/NavBar.vue
```diff
@@ -0,0 +1,77 @@
+<script lang="ts" setup>
+import { BSIcon } from "cdh-vue-lib";
+
+const route = useRoute();
```
XanderVertegaal: Unused variable.

frontend/components/layout/Header.vue
```diff
@@ -0,0 +1,73 @@
...
+        <div class="uu-header-row">
+            <div class="ms-auto">
+                Not logged in
+            </div>
+        </div>
+    </div>
+</template>
+
+<style scoped lang="scss">
+@import "uu-bootstrap/scss/configuration";
```
XanderVertegaal: Are these imports needed? The import in main.scss should ideally import all of uu-bootstrap.

frontend/components/layout/Header.vue
```diff
@@ -0,0 +1,73 @@
+<script setup lang="ts">
```
XanderVertegaal: Having `Header` and `Footer` as single-word component names is in violation of [this ESLint/Vue linting rule](https://eslint.vuejs.org/rules/multi-word-component-names.html), also in the DIAPP. 

Since this is a new project, we could consider calling them `PageHeader` and `PageFooter` or `MRHeader` and `MRFooter`, for instance, or we can disable the ESLint warnings. What do you think?

frontend/package.json
```diff
@@ -12,8 +12,13 @@
     "lint:fix": "eslint . --fix"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^6.1.1",
+    "@fortawesome/free-solid-svg-icons": "^6.1.1",
+    "@fortawesome/vue-fontawesome": "^3.0.1",
     "@nuxt/eslint": "^1.4.1",
     "@nuxt/test-utils": "^3.19.1",
+    "bootstrap": "^5.3.7",
```
XanderVertegaal, line 20: I needed to add uu-bootstrap (with the missing 'v' added to the version name) as an explicit dependency to get the project to work.

(I've opened a PR on Vue-Lib so this can be fixed.)

frontend/nuxt.config.ts
```diff
@@ -27,10 +28,16 @@ export default defineNuxtConfig({
     },
     css: {
       devSourcemap: true,
+      preprocessorOptions: {
+        scss: {
+          api: 'modern-compiler',
+          loadPaths: ['.', 'node_modules'],
+        }
```
XanderVertegaal: Addding this to vite.css.preprocessorOptions.scss removes the hundreds of warnings related to UU-Bootstrap.

```suggestion
    silenceDeprecations: [
    "mixed-decls",
    "function-units",
    "global-builtin",
    "import",
    "color-functions",
    ],
```

XanderVertegaal: We should remove these after UU-Bootstrap is fixed (and perhaps create an issue for this so we won't forget).

frontend/plugins/fontawesome.ts
```diff
@@ -0,0 +1,83 @@
...
+    faCirclePlay,
+    faCircleXmark,
+    faGraduationCap,
+    faArrowRight,
+    faSkull,
+    faCrown
+);
+
+export default defineNuxtPlugin((nuxtApp) => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
XanderVertegaal: Also these comments can go (for me at least). It seems that some old errors have been fixed in the meantime.

XanderVertegaal (approved):
Nice work! Good to see a non-default frontend for MyResearch for the first time!

Fixing the UU-Bootstrap dep in `package.json` is the only thing I would insist on before this PR is merged.

Other nitpicks are in the comments. In addition:
- The favicon doesn't load for me. This isn't a priority, but just so you know.
- The indentation in some files is off. Perhaps you'll want to run `prettier . -w` one last time.

frontend/components/layout/Header.vue
tijmenbaarda: They appear to be needed indeed...

frontend/package.json
tijmenbaarda: Added! I will also create an issue.

frontend/components/layout/Header.vue
tijmenbaarda: I have changed these to `PageHeader` and `PageFooter`.

frontend/plugins/fontawesome.ts
tijmenbaarda: Done!

frontend/nuxt.config.ts
tijmenbaarda: `silenceDeprecation` does not work for me, but I have added `quietDeps: true` to silence warnings from dependencies and I have changed our own imports to `use`, which works fine.

PR #32, opened 2025-07-07
Add workflow for Black and prettier

tijmenbaarda:


EdoStorm96:
I think this mostly looks good ... but I do have some questions/concerns.

First of all the failing check ... We don't have permission to install uu-bootstrap. I am guess this has something to do with what you removed in [1cbef77](https://github.com/CentreForDigitalHumanities/myresearch/pull/32/commits/1cbef77324d1b2819461f27c7a3f6f9af3ae929c)? Can we implement an SSH key in another way? I'm assuming you had a good reason for removing this.

I'm also noting that for the workflows you use yarn, however, I mostly use npm and I think @XanderVertegaal does too? (may be wrong ...) I think it would be good if we make a decision about what package manager we use. I personally do not have a preference, but I think it would be good to be consistent about this.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,59 @@
...
+          cache-dependency-path: "${{ env.WORKING_DIRECTORY }}/package-lock.json"
+      - name: Install dependencies
+        working-directory: ${{ env.WORKING_DIRECTORY }}
+        run: yarn install --frozen-lockfile
+      - name: Check
+        working-directory: ${{ env.WORKING_DIRECTORY }}
+        run: |
+          yarn prettier --version
+          yarn prettier -w . # Change '.' if you don't want to check everything under working-dir
+          git diff | cat
```
XanderVertegaal: I agree with @EdoStorm96 that sticking with NPM is probably for the best (unless you really prefer Yarn).

The following will probably fix this check:

```suggestion
      - name: Install dependencies
        working-directory: ${{ env.WORKING_DIRECTORY }}
        run: npm ci
      - name: Check
        working-directory: ${{ env.WORKING_DIRECTORY }}
        run: |
          npm run prettier -- --version
          npm run prettier -- -w . # Change '.' if you don't want to check everything under working-dir
          git diff | cat
```

XanderVertegaal (changes requested):
Only one change (see comment below). Otherwise this looks good to me! 👍

.github/workflows/pr-checks.yml
tijmenbaarda: Oh sure, yes, I have no idea why I used Yarn here!

tijmenbaarda:
> First of all the failing check ... We don't have permission to install uu-bootstrap. I am guess this has something to do with what you removed in [1cbef77](https://github.com/CentreForDigitalHumanities/myresearch/pull/32/commits/1cbef77324d1b2819461f27c7a3f6f9af3ae929c)? Can we implement an SSH key in another way? I'm assuming you had a good reason for removing this.

Yes, this is still blocking us from merging... `Vue-lib` is dependent on `uu-bootstrap`, and currently it fetches it using `git+ssh`, which requires an SSH key. But since `uu-bootstrap` is a public repository this is not needed. So instead of creating an SSH key I decided to wait for the new release of `Vue-lib` where this problem has been fixed.

XanderVertegaal:
> > First of all the failing check ... We don't have permission to install uu-bootstrap. I am guess this has something to do with what you removed in [1cbef77](https://github.com/CentreForDigitalHumanities/myresearch/pull/32/commits/1cbef77324d1b2819461f27c7a3f6f9af3ae929c)? Can we implement an SSH key in another way? I'm assuming you had a good reason for removing this.
> 
> Yes, this is still blocking us from merging... `Vue-lib` is dependent on `uu-bootstrap`, and currently it fetches it using `git+ssh`, which requires an SSH key. But since `uu-bootstrap` is a public repository this is not needed. So instead of creating an SSH key I decided to wait for the new release of `Vue-lib` where this problem has been fixed.

I'm working on it! 🫡 😉

tijmenbaarda:
> > > First of all the failing check ... We don't have permission to install uu-bootstrap. I am guess this has something to do with what you removed in [1cbef77](https://github.com/CentreForDigitalHumanities/myresearch/pull/32/commits/1cbef77324d1b2819461f27c7a3f6f9af3ae929c)? Can we implement an SSH key in another way? I'm assuming you had a good reason for removing this.
> > 
> > 
> > Yes, this is still blocking us from merging... `Vue-lib` is dependent on `uu-bootstrap`, and currently it fetches it using `git+ssh`, which requires an SSH key. But since `uu-bootstrap` is a public repository this is not needed. So instead of creating an SSH key I decided to wait for the new release of `Vue-lib` where this problem has been fixed.
> 
> I'm working on it! 🫡 😉

Ahahahah yes I know! :)

tijmenbaarda:
With @XanderVertegaal's release of `cdh-vue-lib` everything is working now, so I am merging this.

PR #33, opened 2025-07-09
Implement internationalization through vue-i18n

### LINKED ISSUE #8: Add frontend i18n
Vue-i18n or gettext?

tijmenbaarda:
As discussed with @XanderVertegaal, `vue-gettext` is not a good solution because this package causes problems in combination with the newest Nuxt version and it is not actively maintained. So we will have to go for `vue-i18n`, like we do in DIAPP.

The setup I followed is a little bit different from DIAPP:
- I added the `vue-i18n-extract` package, which automatically extracts strings and adds them to the localization files, and I set up `npm run i18n-extract` to execute this.
- I added a script to check for untranslated strings, which can be run using `npm run i18n-check`. This script can be run using GitHub Actions for automatic checks.
- In line with our decision for the backend, I chose to use the English text as keys instead of tags. According to the vue-i18n documentation, this is supported even though it is not the default use [1]. The keys (in English) are used as a fallback, even if they contain interpolated strings, which means that there is no need to maintain a separate messages file for English.

Note that even though the use of English keys is mentioned in the documentation and string interpolation is supported, I am not 100% sure we will not run into problems when using this. But I think it is worth giving it a try.

Close #8 

[1] https://vue-i18n.intlify.dev/guide/essentials/fallback.html#fallback-interpolation

### DIFF
--- a/frontend/Dockerfile.dev
+++ b/frontend/Dockerfile.dev
@@ -3,6 +3,9 @@ FROM node:22.16.0-alpine
 # Add git.
 RUN apk add --no-cache git
 
+# Add python because of the i18n-check script
+RUN apk add python3
+
 # Set the working directory in the container.
 WORKDIR /app
 
--- a/frontend/Dockerfile.dev
+++ b/frontend/Dockerfile.dev
@@ -1,5 +1,8 @@
 FROM node:22.16.0-alpine
 
+# Add git.
+RUN apk add --no-cache git
+
 # Set the working directory in the container.
 WORKDIR /app
 
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -87,7 +87,7 @@ npm run i18n-extract
 This tool adds missing keys with an initial empty string as values. (A newer version of `vue-i18n-extract` includes
 an option to use `null` as initial value instead, but there is no release for this yet as of July 2025.)
 
-To check whether all keys are available **and** all messages have been translated, run:
+To check whether there are no missing keys **and** all messages have been translated, run:
 
 ```bash
 npm run i18n-check
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -73,3 +73,22 @@ bun run preview
 ```
 
 Check out the [deployment documentation](https://nuxt.com/docs/getting-started/deployment) for more information.
+
+## Internationalization
+
+Internationalization is implemented using the `vue-i18n` package. The message files are in the `locales` subdirectory.
+
+To extract new messages from the source code, run the `vue-i18n-extract` tool:
+
+```bash
+npm run i18n-extract
+```
+
+This tool adds missing keys with an initial empty string as values. (A newer version of `vue-i18n-extract` includes
+an option to use `null` as initial value instead, but there is no release for this yet as of July 2025.)
+
+To check whether all keys are available **and** all messages have been translated, run:
+
+```bash
+npm run i18n-check
+```
--- a/frontend/i18n-check.py
+++ b/frontend/i18n-check.py
@@ -1,0 +1,39 @@
+#!/usr/bin/python3
+import json
+import subprocess
+from pathlib import Path
+
+translation_path = Path(__file__).parent / "locales"
+translation_files = translation_path.glob("*.json")
+
+
+def check_empty_value(dictionary: dict) -> bool:
+    for (key, value) in dictionary.items():
+        if value == "":
+            print(f"Translation for '{key}' is missing.")
+            return True
+        if isinstance(value, dict) and check_empty_value(value):
+            return True
+    return False
+
+
+# First check for missing translation keys
+print("Checking for missing translation keys...")
+results = subprocess.run(
+    "npx vue-i18n-extract report --vueFiles './{components,pages,layouts,composables}/**/*.?(ts|vue)' --languageFiles './locales/*.?(json|yml|yaml|js)'",
+    shell=True,
+    check=True,
+    capture_output=True,
+)
+if "Missing Keys" in results.stdout.decode():
+    print("Missing translation keys found. Run 'npm run i18n-extract' and add the missing translations.")
+    exit(1)
+print("Ok.")
+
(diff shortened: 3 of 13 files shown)


frontend/README.md
```diff
@@ -73,3 +73,22 @@ bun run preview
...
+To extract new messages from the source code, run the `vue-i18n-extract` tool:
+
+```bash
+npm run i18n-extract
+```
+
+This tool adds missing keys with an initial empty string as values. (A newer version of `vue-i18n-extract` includes
+an option to use `null` as initial value instead, but there is no release for this yet as of July 2025.)
+
+To check whether all keys are available **and** all messages have been translated, run:
```
XanderVertegaal: What do you mean with 'are available'?

frontend/package.json
```diff
@@ -9,7 +9,9 @@
     "preview": "nuxt preview",
     "postinstall": "nuxt prepare",
     "lint": "eslint .",
-    "lint:fix": "eslint . --fix"
+    "lint:fix": "eslint . --fix",
+    "i18n-extract": "vue-i18n-extract report --vueFiles './{components,pages,layouts,composables}/**/*.?(ts|vue)' --languageFiles './locales/*.?(json|yml|yaml|js)' --add",
```
XanderVertegaal: My poor Windows machine can't deal with this syntax. I had to adapt it to:

```suggestion
    "i18n-extract": "vue-i18n-extract report --vueFiles \"./{components,pages,layouts,composables}/**/*.{ts,vue}\" --languageFiles \"./locales/**/*.{json,yml,yaml,js}\" --add",
```

Does that work on Linux as well?

frontend/i18n-check.py
```diff
@@ -0,0 +1,39 @@
...
+            return True
+        if isinstance(value, dict) and check_empty_value(value):
+            return True
+    return False
+
+
+# First check for missing translation keys
+print("Checking for missing translation keys...")
+results = subprocess.run(
+    "npx vue-i18n-extract report --vueFiles './{components,pages,layouts,composables}/**/*.?(ts|vue)' --languageFiles './locales/*.?(json|yml|yaml|js)'",
```
XanderVertegaal: Same problem as in package.json:

```suggestion
    'npx vue-i18n-extract report --vueFiles "./{components,pages,layouts,composables}/**/*.{ts,vue}" --languageFiles "./locales/**/*.{json,yml,yaml,js}"',
```

frontend/i18n-check.py
```diff
@@ -0,0 +1,39 @@
...
+    shell=True,
+    check=True,
+    capture_output=True,
```
XanderVertegaal, line 26: Minor optimisation: if you add `text=True`, the decoding happens automatically.

```suggestion
    text=True,
)
if "Missing Keys" in results.stdout:
```

frontend/i18n-check.py
```diff
@@ -0,0 +1,39 @@
...
+    print("Missing translation keys found. Run 'npm run i18n-extract' and add the missing translations.")
+    exit(1)
+print("Ok.")
+
+for translation_file in translation_files:
+    print(f"Checking '{translation_file}' for missing translations...")
+    translations = json.load(open(translation_file, 'r'))
+    if check_empty_value(translations):
+        print("One or more translations are missing.")
+        exit(1)
```
XanderVertegaal, line 38: It's not really necessary but nice to use `with...` here:

```suggestion
    with open(translation_file, "r", encoding="utf-8") as file:
        translations = json.load(file)
        if check_empty_value(translations):
            print("One or more translations are missing.")
            exit(1)
```

frontend/i18n-check.py
XanderVertegaal: Oh, update: with `text=True`, results.stdout may be None, so you get an error: `TypeError: argument of type 'NoneType' is not iterable`.

I'll leave it up to you whether to keep it as it is or change it (and add a null check).

XanderVertegaal:
I've pushed a commit that fixes the Dockerfiles for the frontend container.

XanderVertegaal (changes requested):
Well done! This works really well, and I'm happy with the scripts.

I did have to update the npm commands a little bit to make it work with Windows, so I have made some suggestions in the comments. Please check if they work on Linux as well. If not, we might have to add Windows-only commands (e.g. `i18n-extract:win`).

frontend/package.json
miggol: `docker compose run vue-dev npm run i18n-extract` works on everything :)

I would prefer that we avoid running these scripts outside of the container. That way we can prevent differences in package versions and OSes

miggol: That being said your suggested command does work in the container as well so I guess that for now you can have your cake and eat it

frontend/Dockerfile.dev
```diff
@@ -1,5 +1,8 @@
 FROM node:22.16.0-alpine
 
+# Add git.
+RUN apk add --no-cache git
```
miggol, line 4: Missing `python3` dependency for running i18n-check

miggol (approved):
Just one small request but otherwise approved!

frontend/package.json
XanderVertegaal: I like cake. 🍰

In all seriousness, I like being able to run a dev build outside of Docker as well (just to remove one more layer of abstraction in case of debugging) so I appreciate being able to run this command on my Windows machine as well 😊

tijmenbaarda: OK, so we can keep it as is? :)

frontend/Dockerfile.dev
tijmenbaarda: Good spot!

frontend/README.md
tijmenbaarda: Clarified this!

frontend/package.json
XanderVertegaal: I think @miggol meant that my suggestion works both on Linux and Windows, so it we would use that, that would be great 😇

Issue #6, opened 2025-06-11
Add workflow for Black + Prettier


XanderVertegaal:
Use DIAPP as an example.

tijmenbaarda:
Closed by #32

PR #36, opened 2025-07-14
Feature/graphql

### LINKED ISSUE #16: HelloWorld GraphQL view/page

EdoStorm96:
fixes #16.

So I've made a very simple implementation of graphql for myresearch using graphene for Django and Apollo for vue. Right now, as a proof of concept, I've made a querytype which can get the username and email of all users. This is just a proof of concept and should probably be deleted well before we go into production, but for now I think it's a nice way to test graphql.

This implementation borrows heavily from how things are done in DIAPP.

The graphene side of things can be tested on: http://127.0.0.1:8000/api/graphql
And as a proof-of-concept for getting data from the backend to the frontend, check out: http://127.0.0.1:3000/gql_tst/

@XanderVertegaal helped me out with the final steps. I copied over a lot of config files from DIAPP, which I won't pretend to fully comprehend. But I think this is a good first step for the GraphQL implementation.

Also, we use codegen to (to quote the DIAPP's `working_with_graphql.md`): "automatically creates types for the frontend, so your query results are type-safe." So for now, during development, it's best to run `npm run codegen` in a separate terminal window. Maybe this should be included in the compose file for the dev profile?

Anyway, let me know what you thing. Sorry for the chaotic description. This was quite the challenge for me and I still don't fully understand it all. However, it does what I wanted it to do, so I thought, let's make a PR.


backend/myresearch/settings.py
```diff
@@ -85,7 +98,12 @@
         "PORT": "5432",
     }
 }
-
+# DATABASES = {
```
tijmenbaarda: What happened here? :)

frontend/app.vue
```diff
@@ -5,4 +5,4 @@
     <NuxtLayout>
         <NuxtPage />
     </NuxtLayout>
-</template>
+</template>
```
tijmenbaarda: Somehow many of your files do not end in a newline. In practice this is not really a problem anymore but maybe you can instruct your editor to add a newline at the end of every file.

frontend/generated/gql/fragment-masking.ts
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable */
```
tijmenbaarda: In DIAPP the `generated` directory is added to `.gitignore`, so I think it is not necessary to add these files to git.

tijmenbaarda (approved):
Just a few comments, but my knowledge about GraphQL is limited :)

backend/main/types/UserType.py
```diff
@@ -0,0 +1,21 @@
...
+from django.db.models import QuerySet
+
+from main.models import User
+
+
+class UserType(DjangoObjectType):
+    """
+    NOTE: This Type was written during a very early phase of development as 
+    a proof-of-concept for implementing graphql
+    """
```
XanderVertegaal: I think this note can go. We will need this type in the future.

backend/main/types/UserType.py
```diff
@@ -0,0 +1,21 @@
...
+    class Meta:
+        model = get_user_model()
+        fields = ["username", "email"]
```
XanderVertegaal: Please add "id" to `fields`. It is good practice to add this for every GraphQL type you create that is based on a DB object.

When Apollo (the frontend GraphQL client) receives new data from the backend, it needs to figure out which entities it already has and which ones are new, so it can update the cache accordingly. By default, it will use the `id` property for this, so when you query an object, you will almost always want to query the `id` property as well, even if you do not need it for anything in the frontend.

backend/main/queries.py
```diff
@@ -0,0 +1,13 @@
...
+
+from main.types.UserType import UserType
+from django.contrib.auth import get_user_model
+
+
+class ListUsersQuery(graphene.ObjectType):
+    users = DjangoListField(UserType)
+
+    def resolve_users(self, info):
+        return get_user_model().objects.all()
```
XanderVertegaal: This works for the present development build! 

In practice, however, it is better to use `return UserType.get_queryset(get_user_model.objects, info)` in a query resolver like this. The `get_queryset()` method is where we will inspect the user and do permission checks in the future (cf. DIAPP). (By calling `.objects.all()` on the model directly, you will bypass these checks once we implement them.). I suggest calling `get_queryset()` here too so we get used to consistent/best practices :)

backend/myresearch/settings.py
```diff
@@ -39,18 +42,28 @@
...
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
 ]
 
+CORS_ALLOW_ALL_ORIGINS = True
+CORS_ALLOW_CREDENTIALS = True
+
+CSRF_TRUSTED_ORIGINS = ["http://localhost:3000"]
```
XanderVertegaal, line 66: I have this setting in `local_settings.py` (that doesn't go on the server).

To be honest, I'm not very sure about the best CORS/CSRF settings for our current architecture. @miggol probably has more experience with this, so perhaps we can discuss this during the next PortalDev meeting?

frontend/pages/gql_tst.vue
```diff
@@ -0,0 +1,31 @@
+<script lang="ts" setup>
+import { graphql } from "~/generated/gql";
+import { useQuery } from "@vue/apollo-composable"
```
XanderVertegaal: There are some formatting issues with this file, e.g missing `;` after this line. 

(Not a problem for this PR, but Prettier will definitely complain about this once [#6](https://github.com/CentreForDigitalHumanities/myresearch/issues/6) is merged.)

XanderVertegaal (approved):
Great work! 🍰

I have a few comments to go along with those of Tijmen. Once all of those are addressed, this is good to merge as far as I am concerned. Before deploying, however, we should have a good look at the CSRF/CORS settings. I will make an issue for this.

PR #37, opened 2025-07-18
Attempt to fix eslint config, optionally lint all frontend code

miggol:
I noticed that there was an eslint config and command prepared but that it threw an error. Following this [suggestion](https://typescript-eslint.io/getting-started/typed-linting/) linting now works. Try it yourself:

```
docker compose run vue-dev npm run lint
docker compose run vue-dev npm run lint:fix
```

This advises a buttload of changes and I can't recall if we'd actually agreed on linting the repo, so I haven't committed any of that. Paging our more JS-capable devs to evaluate if the rules are sensible.

Either way I would still like the `npm run lint` command to work for my IDE's sake.

Feel free to commit rule changes or ignores directly to this branch without asking.


XanderVertegaal (approved):
Great find! 🚀 

Most errors seemed to be coming from an inconsistent EOL style. (Windows uses LF+CR; I think Unix systems only use LF nowadays). I've adapted my default line-ending for this project to be LF, so this should not occur in the future (and else ESLint will warn me about it).

The settings as we have now are very strict but I think it's good to try it out. If we hate it, we can always loosen up a bit :)

.git-blame-ignore-revs
XanderVertegaal, line 1: I'd like to know what this does!

tijmenbaarda (approved):
Yes, this works well, with the changes that @XanderVertegaal made!

I am in favour of adding this to the GitHub Actions checks, as long as we don't make it too strict.

We could perhaps remove the Prettier checks, because we have a separate check for that being prepared.

PR #42, opened 2025-07-30
Django-prod and compose Workflows

miggol:
Ready for review! Changes of note:

- New `django-prod` service hosted by gunicorn
- `nginx-prod` variant which serves static files from a shared volume
- Django is now only accessible through URLs prefixed with `/backend/`
- Django tests are now run within the container. I stole the migrations check from the other workflow as well to verify that the container only needs to be built once. 
- Something to be aware of: migrations are now run on every Django container boot

The goal for this PR is to have something we can put on the servers, even if it's just a hello world. So if you think I've missed something to that end (e.g. security-wise) please let me know as well.


backend/myresearch/settings/utils.py
```diff
@@ -0,0 +1,22 @@
+from django.core.exceptions import ImproperlyConfigured
+
+import os
+
+
+def discover(key, default):
```
XanderVertegaal, line 6: Why not simply use `os.getenv()` for this?

backend/Dockerfile
```diff
@@ -3,8 +3,12 @@ FROM python:3.11-alpine
...
+RUN mkdir -p /var/www/backend-static/
 WORKDIR /usr/src/app/backend
-COPY requirements.txt .
-RUN pip install -U pip pip-tools && pip install -r requirements.txt
+COPY . .
 
-COPY . .
\ No newline at end of file
+# Install requirements
+RUN --mount=type=cache,target=/root/.cache/pip
```
XanderVertegaal: Great idea to add this! Shouldn't it be in the same `RUN` statement as `pip install`, however?

backend/myresearch/settings/dev.py
```diff
@@ -0,0 +1,21 @@
+from .utils import discover
```
XanderVertegaal, line 1: Unused import.

compose.yaml
```diff
@@ -6,18 +6,36 @@ services:
       - POSTGRES_PASSWORD=postgres
     healthcheck:
       test: 'psql -c "\l" postgres postgres'
+      start_period: 5s
```
XanderVertegaal, line 9: Really nice! I didn't know this setting existed. 👌

compose.yaml
```diff
@@ -30,19 +48,53 @@ services:
...
+      python manage.py migrate &&
+      python manage.py runserver 0.0.0.0:8000
+      "
     ports:
       - "8000:8000"
     environment:
       DB_HOST: 'mr-postgres'
+      DJANGO_SETTINGS_MODULE: "myresearch.settings.dev"
+
+  django-prod:
```
XanderVertegaal, line 66: For `django-prod` and `nginx-prod` you might want to extend the existing (dev) configs, or vice versa.

In other projects, I'm using the following syntax for this:

```yaml
services:
  app-prod: &app-prod
    container_name: cb-app
    image: cb-app-prod
    profiles: ["prod"]
    ...

  app-dev:
    <<: *app-prod
    image: cb-app-dev
    profiles: ["dev"]
```

All other properties not defined on `app-dev` are taken from `app-prod`. That helps you define shared things like the `hostname` and healthchecks only once.

An example can be found [here](https://github.com/UUDigitalHumanitieslab/coloringbook/blob/develop/docker-compose.yml).

compose.yaml
```diff
@@ -81,6 +133,7 @@ volumes:
   frontend-node-modules:
   pgdata:
   bapyca:
+  basta:
```
XanderVertegaal, line 127: BASTA! 🤌

nginx.conf.prod
```diff
@@ -10,6 +10,15 @@ http {
...
     server {
         listen 80;
+	
+        location /backend/static/ {
+	    alias /var/www/backend_static/;
+	    autoindex off;
+        }
+
+        location /backend/ {
+            proxy_pass http://mr-django:8000;
```
XanderVertegaal, line 23: No `proxy_http_version 1.1` here?

nginx.conf.prod
```diff
@@ -10,6 +10,15 @@ http {
 
     server {
         listen 80;
+	
+        location /backend/static/ {
+	    alias /var/www/backend_static/;
```
XanderVertegaal: Indentation missing :)

nginx.conf.dev
XanderVertegaal, line 1: I'm still not really sure we need separate `.prod` and `.dev` NGINX conf files. Apart from the comments in this file, the only addition in the prod version is the added `location /backend/static/` directive. Would it be a problem to have that in `.dev` as well?

.github/workflows/build-and-test-django.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Compose
+        uses: docker/setup-compose-action@v1
+      - name: Copy test environment variables
+        run: cp .env.example .env
+      - name: Build django-prod container
+        run: docker compose build django-prod
+      - name: Run Django tests
```
XanderVertegaal: You may want to check for missing migrations before running the tests (in the spirit of 'fail early'), in case we end up with an elaborate test suite that takes a while to run only to find out that you're missing a migration.

(This is a moot optimisation suggestion; feel free to ignore!)

backend/myresearch/settings/dev.py
```diff
@@ -0,0 +1,21 @@
...
+
+ALLOWED_HOSTS = [
+    "localhost",
+    "127.0.0.1",
+    "mr-django",
+]
+
+from .generic_settings import *
+
+try:
```
XanderVertegaal, line 25: `generic_settings.py` already imports `.local_settings`, so there is no need to do that here as well.

XanderVertegaal (approved):
Looks good! I don't see any glaring security issues. Have a look at the comments below. None of them are serious enough that they should prevent you from merging.

compose.yaml
miggol: Awesome!

nginx.conf.prod
miggol: I didn't really know why that was in the vue section, I kind of presumed vue preferred 1.1?

I've now read up on it a bit and it seems that HTTP/1.1 should be better in pretty much every case, so I've moved that part up to the `http {}` section.

nginx.conf.prod
miggol: Apparently I used a cocktail of spaces and tabs

.github/workflows/build-and-test-django.yml
miggol: Makes sense!

backend/myresearch/settings/dev.py
miggol: Ah yes, I wanted to remove that from generic settings because local settings have no place in prod. But I forgot

miggol:
Hi! Re-requesting a quick review because this was a pretty big merge that needed some changes. I've set up `NUXT_PUBLIC_API_URL` and modified URLs to point to `/backend/` where necessary.

For me, `graphql-codegen` also did not get installed with the current requirements setup so I added it manually.

PR #43, opened 2025-07-31
Feat/dev data and dockerize codegen

### LINKED ISSUE #23: Beginnings of testdata gen
Preferably with a test to check for models not represented in the test data generator.

### LINKED ISSUE #39: Add codegen to Docker
Parallel to the nuxt dev server, the frontend should run `npm run codegen`.

This can either be done in the frontend container (easiest?) or in a separate container.

EdoStorm96:
So are two fairly unrelated issues in one branch, but they compliment each other nicely.

fixes #23:

So that was my initial plan for this branch, to get the testdata gen started. I pretty much copied the design from the DIAPP test data gen, but removed everything not needed. I also implemented @miggol's suggestion to check for missing models. The way I implemented it now, is that you have to manually update the list `dev_data_models` after you have implemented a new model in the dev data generation. This then get compared against a list which gets generated in `_check_all_models_implemented()`, which gets all models for all `settings.LOCAL_APPS`. This is all working fine. For now we only generate users.

fixes #39:

I also really wanted to automate codegen in the docker setup, so we can use http://localhost:3000/gql_tst/ to see the results of the generated testdata in the frontend. I had to tinker quite a bit with the compose file to get this to work, with generous help from @XanderVertegaal, but it seems to work now. Some things to note:

- the command for vue-dev has been replaced with a .sh file, for easier maintainability.
- As codegen needs the backend to run, vue-dev now depends on django-dev.
- While implementing this condition, I found that django-dev's healthcheck was not working. I now also install curl in the dockerfile to perform the healthcheck.
- As the healthcheck needs a 200 request, we needed to receive this somewhere from the backend. I implemented a blank page for this on http://localhost:8000/healthcheck/ (though using admin for this also works ...) I am not sure if this is bad. It's something I saw online. Let me know what you think of this.

Anyway, it all works, so that is good. Any suggestions for easter eggs in dev_users.json?

### DIFF
--- a/backend/main/management/commands/create_dev_data.py
+++ b/backend/main/management/commands/create_dev_data.py
@@ -6,15 +6,16 @@ from django.apps import apps
 
 from main.models import User
 
+
 class Command(BaseCommand):
     help = "Create dev dataset for myresearch"
 
     # All models for which dev data has been implemented. Add the model to
     # this list when dev data creation has been implemented for this model.
     dev_data_models = [
-            User,
-        ]
-    
+        User,
+    ]
+
     def print(self, options, *args, **kwargs):
         if not options["silent"]:
             print(*args, **kwargs)
@@ -41,12 +42,14 @@ class Command(BaseCommand):
         for fixture in fixtures:
             call_command("loaddata", fixture)
 
-    def _check_all_models_implemented(self,):
+    def _check_all_models_implemented(
+        self,
+    ):
         """
         Gather all models from LOCAL_APPS and check if they are all present in
         dev_data_models.
         """
-        
+
         all_mr_models = []
 
         for app in settings.LOCAL_APPS:
--- a/backend/main/management/commands/create_dev_data.py
+++ b/backend/main/management/commands/create_dev_data.py
@@ -16,6 +16,10 @@ class Command(BaseCommand):
         User,
     ]
 
+    def add_arguments(self, parser):
+        parser.add_argument("--force", action="store_true")
+        parser.add_argument("--silent", action="store_true")
+
     def print(self, options, *args, **kwargs):
         if not options["silent"]:
             print(*args, **kwargs)
--- a/backend/main/management/commands/create_dev_data.py
+++ b/backend/main/management/commands/create_dev_data.py
@@ -1,0 +1,61 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.core.management import call_command
+from django.db import transaction
+from django.apps import apps
+
+from main.models import User
+
+class Command(BaseCommand):
+    help = "Create dev dataset for myresearch"
+
+    # All models for which dev data has been implemented. Add the model to
+    # this list when dev data creation has been implemented for this model.
+    dev_data_models = [
+            User,
+        ]
+    
+    def print(self, options, *args, **kwargs):
+        if not options["silent"]:
+            print(*args, **kwargs)
+
+    def handle(self, *args, **options):
+        if not settings.DEBUG and not options["force"]:
+            raise CommandError(
+                "Refusing to execute command unless DEBUG = True in settings.py"
+            )
+
+        self._check_all_models_implemented()
+
+        with transaction.atomic():
+            self._create_test_users(options)
+
+    def _create_test_users(self, options):
+        """
+        Create mock users for test purposes from fixtures.
+        """
+        fixtures = [
+            "main/management/commands/dev_fixtures/dev_users.json",
+        ]
+
+        for fixture in fixtures:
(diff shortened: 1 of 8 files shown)


backend/main/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,64 @@
...
+
+        all_mr_models = []
+
+        for app in settings.LOCAL_APPS:
+            for mr_model in apps.get_app_config(app).get_models():
+                all_mr_models.append(mr_model)
+
+        for mr_model in all_mr_models:
+            if mr_model not in self.dev_data_models:
+                raise CommandError(
```
XanderVertegaal, line 65: I like this function! I hope you don't mind if I steal this and use it in the DIAPP as well. ;)

I do think a `CommandError` is quite serious, however. A warning would be better in my opinion.

backend/myresearch/settings.py
```diff
@@ -26,29 +26,37 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ["mr-django", ".localhost", "127.0.0.1", "[::1]"]
```
XanderVertegaal, line 29: @miggol 's PR (#42) implements a split between prod/dev settings, so perhaps it's good to wait until that one is merged and then take those changes on board here.

backend/myresearch/urls.py
```diff
@@ -25,4 +26,6 @@
 urlpatterns = [
     path("admin/", admin.site.urls),
     path("api/", include("api.urls")),
+    # Used as a healthcheck by the docker container
```
XanderVertegaal:
```suggestion
    # Used for a healthcheck by the Docker container.
```

compose.yaml
```diff
@@ -66,6 +69,7 @@ services:
     environment:
       # Allows Nuxt to be accessible from the outside.
       - HOST=0.0.0.0
+      - CODEGEN_SCHEMA=http://mr-django:8000/api/graphql
```
XanderVertegaal, line 72: It's great that this is all it takes and that no further changes to `.env` or `codegen.ts` are needed!

backend/myresearch/settings.py
```diff
@@ -26,29 +26,37 @@
...
 
-INSTALLED_APPS = [
-    "django.contrib.admin",
-    "django.contrib.auth",
-    "django.contrib.contenttypes",
-    "django.contrib.sessions",
-    "django.contrib.messages",
-    "django.contrib.staticfiles",
+LOCAL_APPS = [
     # Main local app
```
XanderVertegaal: I don't think this comment is needed.

XanderVertegaal (commented):
The code looks good! 

Given that these issues are completely unrelated, perhaps you could keep them separated in multiple PRs next time :)

I'm running into the same error as you did when trying to build + run the container: "exec ./start_dev.sh: no such file or directory"

Do you remember what you did to fix it? 😅

EdoStorm96:
> Given that these issues are completely unrelated, perhaps you could keep them separated in multiple PRs next time :)

Sorry will do :) 

> I'm running into the same error as you did when trying to build + run the container: "exec ./start_dev.sh: no such file or directory"
> 
> Do you remember what you did to fix it? 😅

No I do not unfortunately ... Nor can I replicate this error. We'll have to look at it together sometime.

I'm also noticing now that when I remove the command `COPY start_dev.sh ./` from frontend/dockerfile.dev that the container still starts up fine for me ...

backend/main/management/commands/create_dev_data.py
EdoStorm96: > I like this function! I hope you don't mind if I steal this and use it in the DIAPP as well. ;)

It'd be an honor ;p

XanderVertegaal:
Thanks for helping me debug this! For future reference: Windows CRLF vs. Linux LF was (part of) the problem.

XanderVertegaal (approved):

backend/main/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,61 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.core.management import call_command
+from django.db import transaction
+from django.apps import apps
+
+from main.models import User
+
+class Command(BaseCommand):
```
XanderVertegaal, line 10: Oh, btw, I think you forgot to copy the arguments here. Based on the rest of this class you'll need at least:

```python
    def add_arguments(self, parser):
        parser.add_argument("--force", action="store_true")
        parser.add_argument("--silent", action="store_true")
```

PR #44, opened 2025-08-01
Feature: dynamic frontend forms (proof-of-concept)

XanderVertegaal:
As demonstrated during the last MyResearch meeting: this is a proof-of-concept for dynamic frontend forms in Vue using UUBootstrap. We can build on this to prepare a mock-up we can use during out next meeting with Simon when discussing the layout.

Everything is based on a config object in `procRegMockConfig.ts`, which mocks data that the backend might provide.

Unsolved / for future PRs (since this one was getting quite large already):
- Styling is still a bit rough; the stepper is not clickable.
- Route parameters: the route currently doesn't change when selecting a (sub)step, but if we want users to share links, we should make this possible.
- Where do we store form data in the frontend? As part of the FormConfig (with separate `value` properties)? Or a separate object altogether?


frontend/nuxt.config.ts
```diff
@@ -30,7 +30,6 @@ export default defineNuxtConfig({
       devSourcemap: true,
       preprocessorOptions: {
         scss: {
-          api: "modern-compiler",
```
XanderVertegaal, line 33: This line produced errors after my Windows 11 reinstall, so I guess a later version of some package made it obsolete?

frontend/composables/useTranslation.ts
```diff
@@ -0,0 +1,22 @@
...
+
+function useTranslateableAttribute<Key extends string>(
+  modelObject: TranslateableObject<Key>,
+  key: Key,
+): string {
+  const currentLocale = i18n.global.locale;
+  const keyNl = `${key}Nl` as keyof TranslateableObject<Key>;
+  const keyEn = `${key}En` as keyof TranslateableObject<Key>;
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
XanderVertegaal: Somehow, `currentLocale.value` is typed as `nl` (a string literal) and I have no idea why.

frontend/eslint.config.mjs
```diff
@@ -47,10 +54,31 @@ export default defineConfigWithVueTs([
...
   {
     ignores: ["node_modules", ".nuxt", ".output", "dist"],
   },
   // prettier
   prettier,
+  {
+    rules: {
+      ...prettier.rules,
+      "prettier/prettier": [
+        "warn",
```
XanderVertegaal: In a future PR, this should be moved to `.prettierrc`.

EdoStorm96 (approved):
Alright, I think this is a very good jumping off point. Nice work! I do not fully comprehend the subtleties of what is going on ... But I understand the gist of it.

I approve but here are my thoughts on your discussion points.

> Styling is still a bit rough; the stepper is not clickable.

Looks pretty good already though!
> Route parameters: the route currently doesn't change when selecting a (sub)step, but if we want users to share links, we should make this possible.

Yes, I think this would be good. I do not find the `selectedStep` / `nextStep` / `previousStep` stystem in MRForm.vue to be very intuitive ... I assume this will get removed once we implement proper routing using slugs? I think this should be done pretty soon, as this will also make adding links to the stepper much easier.
> Where do we store form data in the frontend? As part of the FormConfig (with separate value properties)? Or a separate object altogether?

Again, I am no expert on this stuff, but my intuition says to just store the values in the backend? Then the backend regenerates the formconfig, based on what has just been filled in, whenever a value gets saved. Maybe I am overlooking something or misunderstanding you, but I do not see why values would need to be stored in the frondend?

Nice work anyway. This was some good typescript immersion for me ;p

Hope @tijmenbaarda can also give a second opinion on this :)

frontend/nuxt.config.ts
tijmenbaarda: I hope so... I struggled quite a lot with this part. I am not getting errors now though so I suppose it is all right!

tijmenbaarda (approved):
I have not been able to test this because I got some Apollo-related errors that I am not able to fix anymore today, but the code looks good! I love the way the types are built up.

PR #47, opened 2025-08-14
Doc: create DB script

XanderVertegaal:
Our RSL 'cookie cutter' project template contains a SQL script that helps new users set up a dev database, so I've added it here too. This may save developers (and ourselves, after a clean system install) getting MyResearch to run locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,20 +60,21 @@ local_settings.py file, which contains the following settings:
     CSRF_TRUSTED_ORIGINS = ["http://*:3000"]
     ```
 
-6. Create a new local development database. The Django development server 
-expects a database with the following details. (This can be changed as needed 
-in `settings.py`):
+6. Create a new local development database. The first PostgreSQL start-up will
+normally do this for you, but in case you need to do this manually, follow this
+step. The Django development server expects a database with the following
+details. (This can be changed as needed in `settings.py`):
 - DB name: `myresearch`
 - Host: `localhost`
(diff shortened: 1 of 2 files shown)


create_db.sql
miggol, line 1: Please merge this with `backend/create_db.sql`. Could be placed in either dir, but remember to update the mount in `compose.yaml` to reflect your choice!

README.md
```diff
@@ -61,6 +60,26 @@ local_settings.py file, which contains the following settings:
...
+the database and user with the correct permissions. Run it as follows.
+
+```bash
+psql -U <your-postgres-username> -f create_db.sql
+```
+
+Note that this is not needed if you are using the provided Docker setup. Also, 
+make sure to never use these standard settings in a production environment.
+
+
```
miggol: Does no harm, but the first postgres startup will run `create_db.sql` anyway. So perhaps make it clear that these steps are only necessary if you broke something or are using a nonstandard configuration like sharing your postgres (ew).

miggol (commented):
I think someone got here before you, Xander

XanderVertegaal:
image

I had no idea this file was already there! It makes sense, though, since we have all got working DBs. 😉 I will leave the original file as is, remove mine and simply add some documentation.

README.md
XanderVertegaal: Gross!

PR #48, opened 2025-08-15
Feat/frontpage

### LINKED ISSUE #45: Homepage mockup
Add a title "Welcome to MyResearch!" to the homepage.

Use https://iwcmxf.axshare.com/?id=d81nkn&p=landingspagina_-_empty_state for inspiration.

RoordinkFrank:


frontend/components/AccordionQuestions.vue
```diff
@@ -0,0 +1,63 @@
...
+
+<template>
+  <div class="col-12">
+    <!-- without col-12 accordion will collapse if the body has little content-->
+    <div class="accordion" id="questionsMenu">
+      <div class="accordion-item">
+        <h2 class="accordion-header" id="QuestionOneHeading">
+          <button
+            class="accordion-button collapsed show"
+            type="button"
```
RoordinkFrank: Accordion currently only properly collapses when you close open another accordion item. There is some movement as if it tries to close but is readjusted by something imminently after.
Exact same code works in other places so it feels like something is overriding the behavior.

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,72 @@
...
+  accordionItems.push({
+    id: groupName + i,
+    question: questionsList[i],
+    answer: answersList[i],
+  });
+}
+</script>
+
+<template>
+  <!--  TODO: mw-100 on every div is ugly overriding the (uu-sidebar .div) is better but i have been unable to do that as of now-->
```
RoordinkFrank: mw-100 on every div is ugly overriding the (uu-sidebar .div) is better but i have been unable to do that as of now

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,72 @@
...
+  <!-- TODO: accordion does not always collapse properly, accordion behaviour gets overridden from outside -->
```
RoordinkFrank: accordion does not always collapse properly, accordion behaviour gets overridden from outside.

edit: made a separate issue for this as this may take a while.

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,75 @@
+<script setup lang="ts"></script>
+
+<template>
+  <!-- TODO the first tile is not clickable everywhere, second one changes icon but is also not clickable -->
```
RoordinkFrank: todo: the first tile is not clickable everywhere, second one changes icon but is also not clickable 
tiles really wants div tile directly below it but in that case the nuxtlink isnt the whole tile

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,75 @@
...
+  <!-- tiles really wants div tile directly below it but in that case the nuxtlink isnt the whole tile-->
+
+  <!-- TODO unlike frequentQuestions, recentactivity may need to be hidden once the screen gets smaller -->
```
RoordinkFrank: todo: unlike frequentQuestions, recentactivity may need to be hidden once the screen gets smaller

frontend/pages/index.vue
```diff
@@ -1,17 +1,122 @@
...
+                  Heb het volgende bij de hand: Helemaal Niets
+                </p>
+                <a href="#" class="btn btn-primary">Start Aanvraag</a>
+              </div>
             </div>
+          </div>
         </div>
+      </div>
+      <aside class="uu-sidebar pe-3">
+        <!-- TODO pe-3 does not exactly give 1.25 rem as pixels but it seems very close, low prio -->
```
RoordinkFrank: TODO pe-3 does not exactly give 1.25 rem as pixels but it seems very close, low prio

frontend/shared/mockData.ts
```diff
@@ -0,0 +1,65 @@
...
+const answerTwoProcessingRegister = t(
+  "Omdat ze anders niet in hun schil passen",
+);
+const questionOneEthnicalCommission = t(
+  "Where can i find the documents I have to deliver",
+);
+const answerOneEthnicalCommission = t(
+  "Tijdens de aanvraagprocedure kan er om een of meer bijlagen worden gevraagd. Gebruik daarvoor de juiste (meest recente) voorbeelddocumenten " +
+    useEthicsLinks("model_documents"),
+  //TODO creating a href here is ugly and hard to do, but i still need a link here, give the logic to ethicsLinks?
```
RoordinkFrank: TODO creating a href here is ugly and hard to do, but i still need a link here, give the logic to ethicsLinks?
 still very ugly but it only has to happen one time, also vue refus    questionOneEthnicalCommission,
   questionTwoEthnicalCommission,es <a /> tags?

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,71 @@
+<script setup lang="ts"></script>
+
```
RoordinkFrank: Ethics version for comparison:
image

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+const props = defineProps({
+  questionsGroup: String,
+  title: String,
+});
+
+let groupName = "";
+let questionsList = <string[]>[""];
+let answersList = <string[]>[""];
+
+if (props.questionsGroup == "processingRegister") {
```
EdoStorm96: I think it does not make sense to check which questionGroup has to be rendered here. Instead, I think we should just pas the correct questionGroup as a prop, instead of just passing a string denoting the correct questionGroup.

frontend/shared/mockData.ts
EdoStorm96, line 1: So ideally this data would be structured in a way that makes is directly useable for FrequentQuestions.vue. Now, you still have to wrangle this data quite a bit to get it in a workable place. I'd suggest creating an object type, kindoff like this:
```
interface QuestionAnswer {
  question: string
  answer: string
}

interface QuestionGroup {
  questionGroupName: string;
  questionList: QuestionAnswer[];
}

const ethicsQuestions: QuestionGroup = {
  questionGroupName: "Ethical Comission",
  questionList: [
    {
      question: t(
  "Where can i find the documents I have to deliver",
      ),
      answer: t(
  "Tijdens de aanvraagprocedure kan er om een of meer bijlagen worden gevraagd. Gebruik daarvoor de juiste (meest recente) voorbeelddocumenten " +
    useEthicsLinks("model_documents"),
),
    },
    {
      question: t("How long do responses take"),
      answer: t("Geen idee"),
    },
    // etc ...
  ]
}
```
This then contains all the info that is needed for the frequent questions component, and you do not need to format the data.

(I haven't tested this, so i'm not 100% sure if this will work with the translations ... But I think for formatting the data, this is the direction you should be looking at, as this will be easy to use in your components. Also I think the naming needs to be improved, but that is something you're good at ;p

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,71 @@
...
+          </div>
+          <div class="col-4">25-08-23</div>
+        </div>
+        <p></p>
+        <div class="row justify-content-between">
+          <div class="col-6">TYPE</div>
+          <div class="col-6">Wegwijzer</div>
+        </div>
+        <div class="row justify-content-center">
+          <div class="col-6">Status</div>
```
EdoStorm96: Maybe make this text bold?

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,71 @@
+<script setup lang="ts"></script>
+
+<template>
+  <h2 class="uu-sidebar-header-linked">{{ $t("Recent Activity") }}</h2>
+  <div class="tiles uu-container mw-100">
+    <div class="tile w-100 mw-100">
+      <NuxtLink to="/studies/" class="nav-link">
+        <div class="row justify-content-between">
```
EdoStorm96: Instead of having justify-content-between on many div's, you can wrap all this stuff in a big div and sort out the positioning there.

Also, I think I'd prefer it to be aligned to the left.

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,71 @@
...
+        <div class="row justify-content-start">
+          <div class="col-6">fav color</div>
+          <div class="col-6">Geel</div>
+        </div>
+        <div class="row justify-content-end">
+          <button class="btn btn-arrow-right col-6">Continue</button>
+        </div>
+      </NuxtLink>
+    </div>
+    <a class="tile w-100 mw-100">
```
EdoStorm96: So Ideally, an individual tile would be a separate component, which receives some mockdata too or something. But this is maybe too much to ask right now.

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,71 @@
...
+          <div class="col-6">DRAFT</div>
+        </div>
+        <div class="row">
+          <div class="col-6">fav color</div>
+          <div class="col-6">
+            geel met een hele lange tekst erachter geplakt
+          </div>
+        </div>
+        <div class="row justify-content-end">
+          <button class="btn btn-arrow-right col-6">Continue</button>
```
EdoStorm96: Would be nice if the button has another color, so it gets distinguished from the background.

frontend/composables/useEthicsLinks.ts
EdoStorm96, line 1: this feels like you are reinventing translations. Why not just use the normal translations?

frontend/package.json
EdoStorm96, line 1: This should not have been changed right?

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
+<script setup lang="ts">
```
XanderVertegaal: Very minor nitpick: the order I would keep is `<script lang="ts" setup>`. There is no functional difference here; just a matter of style and consistency :)

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+import { mockData } from "~/shared/mockData";
+
+const props = defineProps({
+  questionsGroup: String,
+  title: String,
+});
+
+let groupName = "";
+let questionsList = <string[]>[""];
```
XanderVertegaal: The syntax with `<...>` is the same as saying `let questionsList = [""] as string[];` This overrides TypeScript's type checker, and it basically says: "I don't want you to look at whatever is in this array. Just treat it as an array of strings." You could write `let questionsList = <string[]>["yes", 78782, true];` and TS would not complain about it, even though it should.

The recommended way to declare an array of strings is `const questionsList: string[] = [""];`. TS will then warn you if you try to put anything else than strings in. (Note also that you could use `const` here, which is preferable to `let`.)

(See also my other comment below about using `computed` here.)

EdoStorm96 (commented):
So I think the page looks pretty good already! I'm quite impressed with what you were able to do! Some things need a bit of tweaking, but that can happen at a later stage. I'm pretty happy with what I see when I visit the page, but I think the code itself can use a bit of work. There is currently a lot of repetition. I understand this at this stage, but I think you could still improve this PR a bit.

In general, I'd like you to think about how a specific component can most easily receive data, so that you do not have to format the data inside the component. So I'd like you to plan ahead what a specific component needs to know, and then create an object which fills the exact needs of this. I've given an example of how I think this should be done for FrequentQuestions.vue / mockData.ts. If you plan ahead how your data will look, the component can in turn become simpler.

If you want, the tiles in RecentActivity can become a separate component for instance, which also receives objects tailored to this need. If you want a little challenge you can try this out, but it is maybe a bit much for now.

So yeah, I like how it looks from the user perspective, but I think the code can be streamlined. Because of my relative lack of expertise in typescript and CSS I'd also like if someone else weighs in on this, before we merge, so therefore I'm just doing this review as a comment.

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
+<script setup lang="ts">
+import { mockData } from "~/shared/mockData";
+
+const props = defineProps({
+  questionsGroup: String,
+  title: String,
+});
```
XanderVertegaal: It's good practice to define the shape of Props and provide it as a type to `defineProps`:

```suggestion
interface Props {
    questionsGroup: string;
    title: string;
}

const props = defineProps<Props>();
```

Note that `string` instead of `String` is used to type strings in TypeScript. Both will work, but [there are subtle differences](https://stackoverflow.com/questions/14727044/what-is-the-difference-between-types-string-and-string). You might want to check the rest of your code for this and make the necessary replacements. :)

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+    id: groupName + i,
+    question: questionsList[i],
+    answer: answersList[i],
+  });
+}
+</script>
+
+<template>
+  <div class="mw-100">
+    <h2 v-if="props.title" class="uu-sidebar-header-linked">
```
XanderVertegaal: If you define props as shown in the comment above, Vue will know that you will always have `title` in props, so you don't have to check for it with `v-if`. Also, as a handy shorthand, you don't have to prepend `props.` to props you want to use in your template. You can directly use:

```html
<h2 class="...">{{ $t(title) }}</h2>
```

frontend/pages/index.vue
```diff
@@ -1,17 +1,132 @@
...
+      <aside class="uu-sidebar pe-3">
+        <FrequentQuestions
+          questionsGroup="ethicalCommission"
+          title="Frequent questions"
+        />
+      </aside>
+    </div>
+    <div class="uu-sidebar-container uu-sidebar-right uu-sidebar-sticky">
+      <div class="uu-sidebar-content">
+        <FrequentQuestions title="Other questions" />
```
XanderVertegaal: This `FrequentQuestions` component does not have the `questionsGroup` attribute set.

frontend/pages/index.vue
```diff
@@ -1,17 +1,132 @@
...
+          {{
+            $t(
+              "At the processing register you register your studies that contain personal data. The processing register is a central place for all studies with personal data. The Dutch government under the Algemene Verordening Gegevensbescherming (AVG) demands that studies are registered where personal data plays a role. Registering such studies also has an advantage for that researcher that registers the study: registration gives a user-friendly overview about aspects in data processing such as the type of personal data, the persons to whom those data apply, retention periods, used software, necessary security measures and the different parties that play a role",
+            )
+          }}
+        </p>
+      </div>
+      <aside class="uu-sidebar pe-3">
+        <FrequentQuestions
+          questionsGroup="processingRegister"
```
XanderVertegaal: It is conventional to use kebab-case in attributes, even for Vue component props, so I'd use `questions-group="processingRegistry"` here. Vue will automatically convert this to camelCased `questionsGroup` inside the component.

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+if (props.questionsGroup == "processingRegister") {
+  groupName = props.questionsGroup;
+  questionsList = mockData.questionsProcessingRegister;
+  answersList = mockData.answersProcessingRegister;
+} else if (props.questionsGroup == "ethicalCommission") {
+  groupName = props.questionsGroup;
+  questionsList = mockData.questionsEthicalCommission;
+  answersList = mockData.answersEthicalCommission;
+}
+const accordionItems = <any>[];
```
XanderVertegaal: Try to avoid using `any` as much as possible, since it disables type checking.

Instead, accordionItems seems to be an array of objects with a fixed shape (always with an `id`, `question` and `answer`), so try to define [a TypeScript interface](https://www.typescriptlang.org/docs/handbook/2/objects.html) for this and type the array accordingly (see my previous comment above).

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
+<script setup lang="ts">
+import { mockData } from "~/shared/mockData";
+
+const props = defineProps({
+  questionsGroup: String,
+  title: String,
+});
+
+let groupName = "";
```
XanderVertegaal: `groupName` seems be given the value of `props.questionsGroup` (and nothing else), so I suggest you use `props.questionsGroup` wherever you want to use `groupName`.

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+let questionsList = <string[]>[""];
+let answersList = <string[]>[""];
```
XanderVertegaal: Since the values of `questionsList` and `answersList` are not constant but depend on props, it is best if you turn them into [Vue `ref` objects](https://vuejs.org/guide/essentials/reactivity-fundamentals.html#declaring-reactive-state-1), using `ref()`. This marks them as 'reactive', so Vue will keep a special eye on them and update the template or other dependent variables if necessary.
In addition, since `accordionItems` is based on `questionsList` and `answerList`, you can define it as a [computed ref](https://vuejs.org/guide/essentials/computed.html#basic-example), like so:

```ts
const accordionItems = computed(() => {
  return questionsList.map((item, index) => ({
    id: groupName + index,
    question: item,
    answer: answersList[index],
  }));
});
```

(I personally prefer `.map` or `.forEach` or the old `for ... in ...` loops, but that more stylistic than functional. :))

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+}
+</script>
+
+<template>
+  <div class="mw-100">
+    <h2 v-if="props.title" class="uu-sidebar-header-linked">
+      {{ $t(props.title) }}
+    </h2>
+    <!-- warning: translations inside this accordion do not change until page is reloaded -->
+    <div class="accordion mw-100" v-bind:id="'accordion' + groupName">
```
XanderVertegaal: Using `v-bind` here is unnecessary. A simple colon (`:`) does the trick just as well.

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+  });
+}
+</script>
+
+<template>
+  <div class="mw-100">
+    <h2 v-if="props.title" class="uu-sidebar-header-linked">
+      {{ $t(props.title) }}
+    </h2>
+    <!-- warning: translations inside this accordion do not change until page is reloaded -->
```
XanderVertegaal: This will probably be fixed if you refactor `questionsList` and `answersList` into `ref`s and `accordionItems` into a `computed`.

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,70 @@
...
+          >
+            <div class="accordion-body mw-100">{{ accordionItem.answer }}</div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped></style>
```
XanderVertegaal: No need for a style tag here, if it's empty.

frontend/composables/useEthicsLinks.ts
tijmenbaarda: Note to self: will look at this @tijmenbaarda

XanderVertegaal (changes requested):
Great job so far! It is clear that you have learned a lot about TypeScript and Vue already. Keep it up and you will be a frontend wizard in no-time! 🚀 

I have only reviewed the `FrequentQuestions` component so far and left a few comments. Have a look and let me know if I can give more context/explanations! After these changes, I will have a closer look at the rest.

frontend/components/FrequentQuestions.vue
RoordinkFrank: Does let questionsList: string[] = [""]; also work in javascript or is it a Typescript only thing? Out of curiosity

XanderVertegaal: `let questionsList = [""];` is valid JavaScript and TypeScript.
`let questionsList: string[] = [""];` is only valid in TypeScript.

frontend/components/AccordionQuestions.vue
RoordinkFrank: Sub-Issue has been added for this.

frontend/shared/mockData.ts
RoordinkFrank: I have done something simular but I ditched types as it's just 2 string in each type, I have used maps instead. Let me know what you think

frontend/components/FrequentQuestions.vue
RoordinkFrank: That is a valid way of doing things but let's say the data comes from a DB in the future. Now you are giving a DB query trough a props. Something I would like to avoid. We don´t know where the final data will comes from yet. Better leave the responsibility to the component.

frontend/components/FrequentQuestions.vue
RoordinkFrank: done

frontend/components/FrequentQuestions.vue
RoordinkFrank: I did things slightly different. I did reuse the map to make question and answers one list. The hint to make it reactive also solved the translations not reloading on language change. Apparently $t(") is already our reactive part we need so computed is no longer needed.

frontend/components/RecentActivity.vue
RoordinkFrank: background has another color now.

frontend/pages/index.vue
```diff
@@ -1,17 +1,146 @@
...
+        </h2>
+        <div>
+          {{
+            $t(
+              "The roadmap is a questionnaire that helps researchers to find out what needs to happen to submit a request to start a research. Do I need to register my research at processing register at all. Perhaps that isn't needed? Questions that the roadmap helps answering",
+            )
+          }}
+          <div class="uu-container">
+            <div class="card-image-left">
+              <img
```
RoordinkFrank: I want something visual here but I had no idea what so yeah...

DjangoCore uses "https://thecatapi.com/". If that matters.

also TODO: Cookie “__Host-SESSION_ID” has been rejected because it is in a cross-site context and its “SameSite” is “Lax” or “Strict”.
That said chances are this is not permanent anyway.

frontend/components/FrequentQuestions.vue
RoordinkFrank: sub-issue made, so documentation remains even if this branch is merged in develop

frontend/pages/index.vue
RoordinkFrank: sub-issue made, so documentation remains even if this branch is merged in develop

frontend/shared/mockData.ts
RoordinkFrank: sub-issue made, so documentation remains even if this branch is merged in develop. This may take a while.

frontend/shared/mockData.ts
```diff
@@ -0,0 +1,72 @@
+const questionOneProcessingRegister = "What is part of personal data?";
```
XanderVertegaal: I think the correct translation for 'Verwerkingsregister' is 'Processing Registry' (ending in '-try'). Also, 'EthnicalCommission' should be changed to 'EthicalCommission' (without '-n-') 😉

frontend/shared/mockData.ts
XanderVertegaal: Kudos for using `Map`s! (They deserve more love, in my opinion.) Still, I think Edo's solution of having static objects instead of building it up incrementally like this is cleaner. You could even refactor it into a single object. The type is not necessary, but recommended.

```ts
type MockQuestions = Record<'processingRegistry' | 'ethicalCommission' | 'other', Record<string, string>>;

const mockQuestions: MockQuestions = {
  processingRegistry: {
    [questionOneProcessingRegistry]: answerOneProcessingRegistry,
    [questionTwoProcessingRegistry]: answerTwoProcessingRegistry,
  },
  ethicalCommission: {
    [questionOneEthicalCommission]: answerOneEthicalCommission,
    [questionTwoEthicalCommission]: answerTwoEthicalCommission,
    [questionThreeEthicalCommission]: answerThreeEthicalCommission,
    [questionFourEthicalCommission]: answerFourEthicalCommission,
  },
  other: {
    "When will the Ethical Commission be added to to My Research?": "After an estimated 2000 cups of coffee.",
    "Why am I looking at a picture of a cat? ": "https://cataas.com/cat",
  }
};
```

Even better would be to dispense with the variables and use strings directly:

```ts
const mockQuestions: MockQuestions = {
  processingRegistry: {
    "What is part of personal data?": "Personal data is dat that tells something about a living person of whom you know the identity, from whom you can discern the identity or whom are otherwise recognizable. For example: contact information is personal data, but also a lot of research data is personal data. Measurement data, answers on a questionnaire, or interviews, observations, demographic data and much more. Data that can not be traced back to personal data are anonymous data, those are not personal data",
    "Why are bananas curved?": "Because they are yellow",
    ...
  },
  ...
};
```

I think this would be easier to read/use and involves less code.

frontend/shared/mockData.ts
```diff
@@ -0,0 +1,72 @@
...
+  "Why am I looking at a picture of a cat? ",
+  "https://cataas.com/cat", //unable to display cat links in accordion, a disappointment.
+);
+
+let questionAnswers = new Map<string, Map<string, string>>();
+questionAnswers.set("processingRegister", processingQuestionAnswers);
+questionAnswers.set("ethicalCommission", ethicalCommissionQuestionAnswers);
+questionAnswers.set("other", otherQuestionAnswers);
+
+export class mockData {
```
XanderVertegaal: Classes in JS always start with a capital: `MockData`.

frontend/shared/mockData.ts
XanderVertegaal: An added benefit is that you can simply access `mockQuestions.ethicalCommission` without having to call `.get()` and deal with potential `undefined`s.

frontend/shared/mockData.ts
```diff
@@ -0,0 +1,72 @@
...
+);
+
+let otherQuestionAnswers = new Map<string, string>();
+otherQuestionAnswers.set(
+  "When will the Ethical Commission be added to to My Research?",
+  "After an estimated 2000 cups of coffee.",
+);
+otherQuestionAnswers.set(
+  "Why am I looking at a picture of a cat? ",
+  "https://cataas.com/cat", //unable to display cat links in accordion, a disappointment.
```
XanderVertegaal: 😞😞😞

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,47 @@
...
+</template>
+
+<style scoped>
+.tiles .tile {
+  aspect-ratio: 2/1;
+}
+.tiles .tile:hover {
+  background: white;
+}
+</style>
```
XanderVertegaal: This is where SCSS could come in handy:

```scss
<style lang="scss" scoped>
.tiles .tile {
  aspect-ratio: 2/1;

  &:hover {
    background: white;
  }
}
</style>
```

As you see, it allows you to nest declarations: the `&` is expanded to the parent selector (`.tiles .tile`) when the SCSS is compiled to CSS.

frontend/components/RecentActivity.vue
XanderVertegaal: Looks pretty close! Good job!

frontend/composables/useEthicsLinks.ts
```diff
@@ -0,0 +1,36 @@
...
+      }
+      return ethicsLinks[i].en;
+    }
+  }
+  return "";
+}
+
+const ethicsLinks = [
+  {
+    tag: "FEtC_H",
```
XanderVertegaal: You can refactor this into an object and use the tag as the key, so you don't have to traverse the whole list.

```ts
type AcceptedTag = 'FEtC_H' | 'FEtC_H_regulations' | 'model_documents';

const ethicsLinks: Record<AcceptedTag, {nl: string; en: string }> = {
    "FEtC_H": {
        nl: "https://fetc-gw.wp.hum.uu.nl/",
        en: "https://fetc-gw.wp.hum.uu.nl/en/",
    },
    "FEtC_H_regulations": {
        nl: "https://fetc-gw.wp.hum.uu.nl/reglement-fetc-gw/",
        en: "https://fetc-gw.wp.hum.uu.nl/en/regulations-fetc-h/",
    },
    "model_documents": {
        nl: "https://intranet.uu.nl/documenten-ethische-toetsingscommissie-gw",
        en: "https://intranet.uu.nl/en/documents-ethics-assessment-committee-humanities",
    },
}
```

You can then simplify the function above to:

```ts
export default function useEthicsLinks(tag: AcceptedTag): string {
  const { locale } = i18n.global;

  if (locale === 'nl') {
    return ethicsLinks[tag].nl;
  } else if (locale === 'en') {
    return ethicsLinks[tag].en;
  } else {
    return "";
  }
}
```

frontend/components/FrequentQuestions.vue
```diff
@@ -0,0 +1,57 @@
...
+        v-for="(questionAnswer, index) in questionAnswers"
+        :key="questionsGroup + index"
+      >
+        <div class="accordion-item mw-100">
+          <h2 class="accordion-header">
+            <button
+              class="accordion-button"
+              type="button"
+              data-bs-toggle="collapse"
+              v-bind:data-bs-target="'#collapse' + questionsGroup + index"
```
XanderVertegaal: Please also remove `v-bind` elsewhere in this component.


```suggestion
              :data-bs-target="'#collapse' + questionsGroup + index"
```

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,47 @@
+<script setup lang="ts"></script>
+
+<template>
+  <h2 class="uu-sidebar-header-linked">{{ $t("Recent Activity") }}</h2>
+  <div class="tiles uu-container mw-100">
+    <a class="tile w-100 mw-100 justify-content-center">
+      <NuxtLink to="/studies/" class="nav-link">
+        <div class="row">
+          <div class="col-9">
+            <strong>refnumber + Titelmet langere naam</strong>
```
XanderVertegaal:
```suggestion
            <strong>refnumber + Titel met langere naam</strong>
```

frontend/pages/index.vue
```diff
@@ -1,17 +1,155 @@
...
             </div>
+          </div>
         </div>
+      </div>
+      <aside class="uu-sidebar pe-3">
+        <RecentActivity />
+      </aside>
     </div>
+    <div class="uu-sidebar-container uu-sidebar-right uu-sidebar-sticky">
+      <div class="uu-sidebar-content">
```
XanderVertegaal: Also here, I think it's `Processing Registry`, not `-er`.

frontend/pages/index.vue
```diff
@@ -1,17 +1,155 @@
...
+            $t(
+              "Please Note: Accepted applications go to the archive. See above menu. The archive is available for everyone at the same facility with with a Solid-ID. ",
+            )
+          }}
+        </p>
+        <p>
+          <a :href="useEthicsLinks('FEtC_H')" target="_blank">{{
+            $t("UU-webpage of the FEtC-H")
+          }}</a>
+        </p>
```
XanderVertegaal:
```suggestion
              "During the application it may be possible that you need submit certain documents. Use the most recent model documents.",
```

frontend/pages/index.vue
```diff
@@ -1,17 +1,155 @@
...
+              "For questions about the processing register contact the privacy officer: ",
+            )
+          }}<a href="mailto:l.j.w.m.kessels@uu.nl">l.j.w.m.kessels@uu.nl</a>
+        </p>
+        <p>
+          {{ $t("For compliments contact: ")
+          }}<a href="mailto:portaldev.gw@uu.nl">portaldev.gw@uu.nl</a>
+        </p>
+      </aside>
+    </div>
```
XanderVertegaal: SCSS can help here:

```scss
.card-image-left {
    display: flex; // Or use the class `d-flex` in the template.

    & img {
        width: 30%;
    }
}
```

frontend/locales/nl.json
```diff
@@ -5,5 +5,46 @@
...
+  "Welcome!": "Welkom!",
+  "Home": "Home",
+  "What is MyResearch?": "Wat is MyResearch?",
+  "Roadmap": "Wegwijzer",
+  "Processing register": "Verwerkingsregister",
+  "Ethical commission": "Ethische commissie",
+  "Recent Activity": "Recente Activiteit",
+  "Frequent questions": "Veelgestelde vragen",
+  "Other questions": "Overige vragen",
+  "My Research is a website that helps researchers with submitting a request to start a research. On this website you can register your research at the processing register. But first you should fill in the roadmap": "My research is een website dat onderzoekers helpt met het aanvragen van onderzoek. Op deze website kan uw onderzoek worden geregistreerd bij het verwerkingsregister. Maar eerst is het de bedoeling dat een wegwijzer wordt ingevuld.",
```
```suggestion
  "My Research is a website that helps researchers with submitting a request to start a research. On this website you can register your research at the processing registry. First, however, you should fill in the roadmap": "My research is een website die onderzoekers helpt met het aanvragen van onderzoek. Op deze website kan uw onderzoek worden geregistreerd bij het verwerkingsregister. Eerst is het echter de bedoeling dat een wegwijzer wordt ingevuld.",
```

frontend/locales/nl.json
```diff
@@ -5,5 +5,46 @@
...
+  "The roadmap is a questionnaire that helps researchers to find out what needs to happen to submit a request to start a research. Do I need to register my research at processing register at all. Perhaps that isn't needed? Questions that the roadmap helps answering": "De wegwijzer is een vragenlijst dat onderzoekers helpt uit te vogelen wat er moet gebeuren om een onderzoek aan te vragen. Moet ik voor mijn onderzoek nou wel of niet een mijn gegevens registeren in het verwerkingsregister? Mogelijk hoeft het niet eens. Iets wat de wegwijzer samen met u uitzoekt. ",
```
```suggestion
  "The Roadmap is a questionnaire that helps researchers to find out what needs to happen before starting a research project. Do I need to register my research at processing registry at all or not? The Roadmap helps you answer this question.": "De Wegwijzer is een vragenlijst die onderzoekers helpt om erachter te komen wat er moet gebeuren voordat een onderzoeksproject van start kan gaan. Moet ik voor mijn onderzoek mijn gegevens registeren in het verwerkingsregister? Mogelijk hoeft het niet eens. U komt erachter door de Wegwijzer in te vullen.",
```

frontend/locales/nl.json
```diff
@@ -5,5 +5,46 @@
...
+  "Start a new roadmap": "Start een nieuwe wegwijzer",
```
XanderVertegaal, line 20: We only have one Roadmap, right?

```suggestion
  "Start the Roadmap": "Start de Wegwijzer",
```

frontend/locales/nl.json
```diff
@@ -5,5 +5,46 @@
...
+  "After an estimated 2000 cups of coffee.": "Na ongeveer 2000 kopjes koffie.",
+  "During the application it may be possible that you need submit certain documents. Use the most recent model documents. ": "Tijdens de aanvraagprocedure kan er om een of meer bijlagen worden gevraagd. Gebruik daarvoor de juiste (meest recente) voorbeelddocumenten.",
+  "Where can i find the documents I have to deliver" : "Waar kan ik de documenten vinden die ik moet opleveren?",
+  "How long do responses take after submission?" : "Hoelang duurt het voordat ik een reactie krijg na het indienen?",
+  "Help my draft application suddenly changed" : "Help mijn draft applicatie is opeens veranderd",
+  "It is possible for the supervisor to change your application. " : "Het is mogelijk dat een supervisor je applicatie heeft veranderd",
+  "What happens if my application gets rejected? " : "Wat gebeurd er als mijn applicatie afgewezen wordt?",
+  "You will need to make a revision and resubmit the application. " : "U moet een revisie maken en opnieuw de applicatie indienen. ",
+  "Why am I looking at a picture of a cat? " : "Waarom kijk ik naar een foto van een kat? ",
```
XanderVertegaal: Why is this a translated string?

frontend/pages/index.vue
XanderVertegaal: Well done! And don't worry about the error. We will probably (sadly) have to remove the cat picture at some point anyway (or hide it).

XanderVertegaal (approved):
Well done! We now have a front page with some actual information on it (and a cat! <3)

I'd like you to take another look at the definition of the mock data (as suggested by Edo and me) and check the names of the Processing Registry and the Ethical Committee. Otherwise, I think this is good to go! The translations/texts can be edited/improved in another PR. (I made a few suggestions here already.)

frontend/locales/nl.json
RoordinkFrank: We are saving one roadmap per research no? The questions stay the same but a researcher can have multiple roadmap results.

frontend/locales/nl.json
RoordinkFrank: Mostly to prevent console warnings once it's to develop. I don´t know how long certain mockdata will stay.

frontend/shared/mockData.ts
RoordinkFrank: Implemented the nested map is indeed a bit much.

frontend/locales/nl.json
RoordinkFrank: Very useful feedback

frontend/locales/nl.json
RoordinkFrank: done.

frontend/pages/index.vue
RoordinkFrank: I think this is outdated. This is now part of frequentQuestion

frontend/composables/useEthicsLinks.ts
tijmenbaarda: This was different than I thought - I agree with @EdoStorm96 that normal translations fit better here.

RoordinkFrank: useEthicsLinks.ts is removed and link translations are now in the translationsfile.

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,48 @@
...
+  <div class="tiles uu-container mw-100">
+    <a class="tile w-100 mw-100 justify-content-center">
+      <NuxtLink to="/studies/" class="nav-link">
+        <div class="row">
+          <div class="col-9">
+            <strong>refnumber + Titel met langere naam</strong>
+          </div>
+          <div class="col-3">25-08-23</div>
+        </div>
+        <p></p>
```
tijmenbaarda: Avoid an empty paragraph for additional spacing; instead you can use the `mb-3` class (or some other number) on the row above.

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,48 @@
+<script setup lang="ts"></script>
+
+<template>
+  <h2 class="uu-sidebar-header-linked">{{ $t("Recent Activity") }}</h2>
+  <div class="tiles uu-container mw-100">
+    <a class="tile w-100 mw-100 justify-content-center">
```
tijmenbaarda: This `a` element is not needed here: directly inside you put a `NuxtLink` which is converted into an `a` element as well, so you get two `a` elements inside each other, which is incorrect. You can change this one by a `div`.

Visually it does not matter, but you can see that the outer part of the tile is currently clickable and it leads to nowhere.

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,48 @@
...
+        </div>
+        <div class="row">
+          <div class="col-6">fav color</div>
+          <div class="col-6">
+            geel met een hele lange tekst erachter geplakt
+          </div>
+        </div>
+        <p></p>
+        <div class="row justify-content-end">
+          <button class="btn btn-primary btn-arrow-right col-12">
```
tijmenbaarda: I would suggest to create the link at this point rather than making the whole tile into a link. Having an `a` element with multiple elements inside is not formally incorrect, but it is advised against for accessibility reasons:
https://govtnz.github.io/web-a11y-guidance/wct/cards/#making-the-whole-card-interactive

What you could do instead is just make the button clickable by moving the `NuxtLink` here and just giving it the same button classes (the same way as it was done in FETC).

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,48 @@
...
+          <div class="col-6">DRAFT</div>
+        </div>
+        <div class="row">
+          <div class="col-6">fav color</div>
+          <div class="col-6">
+            geel met een hele lange tekst erachter geplakt
+          </div>
+        </div>
+        <p></p>
+        <div class="row justify-content-end">
```
tijmenbaarda: You have a Bootstrap `row` here but the column is defined by putting the button element directly under it. Maybe that works, but it is more common to put the columns in separate rows.

But you don't need a row here, because there is only one element to put inside. Instead you can use a `div` with as its classes `d-flex justify-content-end`. That way the button will indeed appear on the right and be smaller, so that the arrow (which turns black when hovering) is placed at the right part of the button.

frontend/components/RecentActivity.vue
```diff
@@ -0,0 +1,48 @@
...
+          </button>
+        </div>
+      </NuxtLink>
+    </a>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+.tiles .tile {
+  aspect-ratio: 2/1;
```
tijmenbaarda: Instead of `2/1`, you could also use `unset`.

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,50 @@
+<script setup lang="ts">
+import springURL from "@/assets/images/covers/coverimage-spring.jpg";
+import summerURL from "@/assets/images/covers/coverimage-summer.jpg";
+import autumnURL from "@/assets/images/covers/coverimage-autumn.jpg";
+import winterURL from "@/assets/images/covers/coverimage-winter.jpg";
+
+const seasons = {
+  // originally the images that come from Ethics had CSS.classes as well,
```
tijmenbaarda: This comment can be removed

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,50 @@
...
+
+<template>
+  <div class="uu-cover h-400">
+    <img :src="seasonCover.image" class="uu-cover-image w-100" alt="cover" />
+    <div class="uu-cover-copyright uu-cover-copyright-right">
+      {{ $t("Cover by") }} {{ seasonCover.author }}
+    </div>
+  </div>
+</template>
+<style scoped></style>
```
tijmenbaarda: Can be removed

frontend/pages/index.vue
```diff
@@ -1,17 +1,157 @@
...
+        <p>
+          <a
+            :href="$t('https://fetc-gw.wp.hum.uu.nl/en/regulations-fetc-h/')"
+            target="_blank"
+            >{{ $t("Regulations of the FEtC-H") }}</a
+          >
+        </p>
+        <p>
+          {{
+            $t(
```
tijmenbaarda: I would remove the lorem ipsum, because the index page now contains enough real text!

tijmenbaarda (changes requested):
Good work!

I have some comments on the recent activity part. All other comments have been addressed, so after this the PR is ready to merge!

frontend/pages/index.vue
RoordinkFrank: done.

frontend/components/SeasonalCover.vue
RoordinkFrank: done

frontend/components/SeasonalCover.vue
RoordinkFrank: done

frontend/components/RecentActivity.vue
RoordinkFrank: image

RoordinkFrank: I can see where you are coming from, the visual part is indeed a problem. I guess this also has to do with that I am unsure if the button or the whole tile should be clickable.

frontend/components/RecentActivity.vue
RoordinkFrank: It's entirely removed with card implementation so no longer relevant.

frontend/components/RecentActivity.vue
RoordinkFrank: I copied something simular from Ethicc in the new card implementation, it uses mt-3

frontend/components/RecentActivity.vue
RoordinkFrank: Only the button is a link now.

frontend/components/RecentActivity.vue
RoordinkFrank: After some discussion the whole tile will be removed and replaced with the Ethics implementation as base.

frontend/components/RecentActivity.vue
RoordinkFrank: Removed.

tijmenbaarda (approved):

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -0,0 +1,109 @@
...
+          <span v-if="activity.status == 'SUBMITTED_TO_SUPERVISOR'">{{
+            $t("Assess")
+          }}</span>
+        </NuxtLink>
+      </div>
+    </div>
+    <div v-if="isRoadmap(activity)" class="card-body mw-100">
+      <h3 class="card-title">{{ activity.refNumber }}</h3>
+      <div>{{ activity.roadmapBody }}</div>
+      <div class="d-flex mt-3 mw-100">
```
RoordinkFrank: Buttonsize can change depending on the text content. It doesn´t look that great anymore in my opinion as the size difference is quite big. I could make all buttons have a Ga verder but that isn´t always true.

Perhaps a different wording  for View Conclusion that is shorter?

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -0,0 +1,109 @@
...
+  refNumber: 3_2025,
+  roadmapBody: "body content of roadmap",
+};
+
+const recentActivity: RecentActivity[] = [
+  staticDraftProposal,
+  staticSendToProposal,
+  staticRoadmap,
+];
+type RecentActivity = Proposal | Roadmap;
```
RoordinkFrank, line 51: type RecentActivity is not a great name but I have no idea what else to name it.

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -0,0 +1,109 @@
...
+      <div class="d-flex mt-3 mw-100">
+        <NuxtLink href="#" class="ms-auto btn btn-primary btn-arrow-right">
+          <span v-if="activity.status == 'DRAFT'">{{ $t("Continue") }}</span>
+          <span v-if="activity.status == 'SUBMITTED_TO_SUPERVISOR'">{{
+            $t("Assess")
+          }}</span>
+        </NuxtLink>
+      </div>
+    </div>
+    <div v-if="isRoadmap(activity)" class="card-body mw-100">
```
RoordinkFrank: The div that has the v-if should always have a card-body class here. As it's done two times it is a bit redundant. Any way to make this cleaner?
Something similar goes for button, uses the same .css classes.

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -0,0 +1,109 @@
...
+  dateSubmitted: "2025-01-24",
+  lastEdited: "2025-04-23",
+};
+const staticRoadmap: Roadmap = {
+  refNumber: 3_2025,
+  roadmapBody: "body content of roadmap",
+};
+
+const recentActivity: RecentActivity[] = [
+  staticDraftProposal,
```
RoordinkFrank, line 47: I will change this to two recentActivities before merge, roadmap doesn´t have enough text to justify putting more in. For now it's easier to review and refactor this way.

frontend/components/frontpage/RecentActivity.vue
EdoStorm96: I suppose you would just have the div with classes `card-body mw- 100` and move the v-if statements down one level. So then for every activity it would create this div which is always needed, and you only start differentiating one the level of div's that are actually different. Hope that makes sense ;p

EdoStorm96 (approved):
Nice to see that this has improved a lot :)

For me, I would prefer some more modern icons ... Maybe the ones that @miggol suggested on Teams?

Anyway, this is all a matter of preference. I think this is looking good for now!

Ps. I am getting this error when I load the page:
```
 [vue-tsc] File '/app/composables/useEthicsLinks.ts' not found. The file is in the program because: Matched by include pattern '.nuxt/../**/*' in '/app/tsconfig.json' 
```
Do you know what causes this?

frontend/components/frontpage/RecentActivity.vue
RoordinkFrank: It does, does not give the solution for buttons but it's better to only do the if statement from where itś needed.

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -0,0 +1,132 @@
...
+          <div class="mw-100">{{ $t("Type") }}: {{ activity.type }}</div>
+          <div class="mw-100">
+            {{ $t("Status") }}:
+            <img
+              class="icon"
+              v-if="activity.status == 'DRAFT'"
+              :src="pencil"
+              alt="pencil.png"
+            />
+            <span v-if="activity.status == 'DRAFT'"
```
RoordinkFrank, line 85: Both are side by side for comparison.

PR #49, opened 2025-08-20
Feature/form models

XanderVertegaal:
This PR adds:
- a new app 'form' to the backend;
- models for the various form elements (form, steps, questions, answers) and side information;
- GraphQL types for the models related to forms and the questions;
- a query that retrieves the most recently created form (for testing);
- frontend components to display some of these question types (DateQuestion, TextQuestion, SelectQuestion). Others remain to be built, but this should at least show a proper form.
- updates to the `create_dev_data.py` script created by @EdoStorm96 to generate a full form with steps, substeps, questions and side information. Run the management command to create a form!
- two new dependencies for the creation of development data:
  - tqdm provides CLI loading bars (similar to the ones hand-made by Ty for DIAPP)
  - faker provides lorem-ipsum like fake content in multiple locales

The various models used for answers are up for discussion. They are very small at the moment, the only difference being the type of their 'value' field. As an alternative to a lot of very small models, we can have one sparse 'Answer' model that has fields for every question type, and all but one would be empty, which is also not very nice. Presently I do not have a preference for either.

image


backend/form/migrations/0001_initial.py
XanderVertegaal, line 1: TIL that django-modeltranslations creates nullable fields by default, which is not what I want. There is no way to change this, apparently. (There is a global setting to make these fields 'blankable', however.)

backend/api/graphql/schema.py
```diff
@@ -1,10 +1,31 @@
...
-class Query(ListUsersQuery, ObjectType):
+class Query(ListUsersQuery, FormQueries, ObjectType):
     pass
 
 
-schema = Schema(query=Query)
+schema = Schema(
+    query=Query,
+    types=[
+        BaseQuestionInterface,
```
XanderVertegaal, line 28: We are not querying any of these types directly in our current queries. For this reason we need to manually add them to our GraphQL schema before we can use them in our queries and types.

frontend/components/layout/NavBar.vue
```diff
@@ -33,11 +33,7 @@ import { BSIcon } from "cdh-vue-lib";
                         </NuxtLink>
                     </li>
                     <li>
-                        <NuxtLink
-                            to="/studies"
-                            class="nav-link"
-                            active-class="active"
-                        >
+                        <NuxtLink to="/" class="nav-link" active-class="active">
```
XanderVertegaal, line 59: This simply removes console warnings. Obviously we will want to put '/studies' back at some point.

frontend/components/shared/FormSideBar.vue
```diff
@@ -1,28 +1,37 @@
...
-        link: string;
-    }[];
-    extraInfo?: {
-        textNl: string;
-        textEn: string;
-    }[];
-}
+import { graphql, useFragment, type FragmentType } from "~/generated/gql";
+
+const StepInfoFragment = graphql(`
```
XanderVertegaal: Fragments in Apollo are neat in that they let you define the data that you need in the component that needs it. Any parent component doing the actual query can incorporate this fragment and then pass an object containing the requested data down to this component.

frontend/pages/proc-reg/index.vue
```diff
@@ -1,7 +1,89 @@
...
+const GET_FORM = graphql(`
+    query GetForm {
+        form {
+            id
+            nameEn
+            nameNl
+            descriptionEn
+            descriptionNl
+            steps {
+                id
```
XanderVertegaal: The fields requested for `steps` are almost identical to those requested for `substeps`. In the future, this should be refactored to a reuseable fragment, but I could not get this to work properly now.

backend/form/types/QuestionType.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @classmethod
+    def get_queryset(
+        cls, queryset: QuerySet[TrueFalseQuestion], info: ResolveInfo
+    ) -> QuerySet[TrueFalseQuestion]:
+        return queryset
+
+
+# Union Type
+class QuestionType(Union):
```
XanderVertegaal, line 133: This is very neat: GraphQL lets you define a Union type that consists of multiple subtypes. At the same time, you can define shared fields using an Interface.

You can query shared and subtype-specific fields in the following way: ❤️ 

```graphql
questions {
    id
    textEn
    textNl
    descriptionEn
    descriptionNl
    required
    ... on SelectQuestionType {
        options {
            id
            labelNl
            labelEn
            defaultSelected
        }
    }
    ... on TextQuestionType {
        placeholderNl
        placeholderEn
        lines
    }
}
```

backend/form/models.py
```diff
@@ -0,0 +1,231 @@
...
+        blank=True,
+        help_text="The order of this step within its parent step.",
+    )
+
+    class Meta:
+        constraints = [
+            models.CheckConstraint(
+                check=Q(parent_order__isnull=False, form_order__isnull=True)
+                | Q(parent_order__isnull=True, form_order__isnull=False),
+                name="either_form_order_or_parent_order",
```
EdoStorm96: I'm not sure if this is complete ... I think it would be nice to ensure that we do not have duplicate orders and to make sure a step gets the correct order ... I am thinking for the logic:
- If a step has a parent, it should have a parent_order
- If a step does not have a parent, it should have a form_order.
- If a step has no parent, ensure its form_order is unique among other steps with no parent of the same form/formconfig.
- If a step has a parent, ensure its order is unique among other steps with the same parent.

What do you think?

backend/form/models.py
```diff
@@ -0,0 +1,231 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+
+class Form(models.Model):
+    name = models.CharField(max_length=200)
+    description = models.TextField(null=True, blank=True)
+    config = models.OneToOneField(
+        FormConfig, on_delete=models.CASCADE, related_name="form"
+    )
```
EdoStorm96: I am not sure I see the added value of having a separate Form and Formconfig right now ... Can't we just cut out the middle man (`Form`) in this case?

backend/form/types/QuestionType.py
EdoStorm96: Fancy stuff!

backend/form/types/QuestionType.py
```diff
@@ -0,0 +1,165 @@
...
+    text_en = String(required=True)
+    description_nl = String(required=True)
+    description_en = String(required=True)
+    required = Boolean(required=True)
+
+
+class DateQuestionType(DjangoObjectType):
+    class Meta:
+        model = DateQuestion
+        interfaces = [BaseQuestionInterface]
```
EdoStorm96, line 36: So I gather this also adds the fields from BaseQuestionInterface to a type? I do not fully understand how this works ... Because `id` is in fields and in interfaces?

backend/main/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,228 @@
...
+            DateQuestion.objects.create(
+                **_base_question_fields(step), future_only=self.faker.pybool()
+            )
+
+        def _create_number_question(step: Step) -> None:
+            NumberQuestion.objects.create(
+                **_base_question_fields(step), positive_only=self.faker.pybool()
+            )
+
+        def _create_file_upload_question(step: Step) -> None:
```
EdoStorm96: I'd prefer if question types which are not yet implemented in the frontend, are not yet loaded into the backend. I was pretty confused during this review about why some questions were not appearing.

backend/main/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,228 @@
...
+MAX_STEPS_PER_FORM = 10
+MIN_SUBSTEPS_PER_STEP = 0
+MAX_SUBSTEPS_PER_STEP = 5
+MIN_QUESTIONS_PER_STEP = 1
+MAX_QUESTIONS_PER_STEP = 5
+
+ALL_QUESTIONS = ["select", "text", "true_false", "date", "number", "file_upload"]
+
+
+class Command(BaseCommand):
```
EdoStorm96, line 45: It'd be nice if you could merge the version of this command on develop into here, which also adds users and checks if all models have been accounted for.

frontend/components/shared/FormSideBar.vue
EdoStorm96: Oof snazzy (but makes my head spin for now ;p)

frontend/components/shared/FormStep.vue
```diff
@@ -1,37 +1,54 @@
...
+                <SharedSelectQuestion
+                    v-else-if="isSelectQuestion(question)"
                     :question="question"
                 />
-                <DateQuestion
-                    v-else-if="question.type === 'date'"
+                <SharedDateQuestion
+                    v-else-if="isDateQuestion(question)"
                     :question="question"
                 />
```
EdoStorm96: Maybe for now I'd like something like:
```
                <p v-else>{{ question.__typename }} does not have a component yet. 😢</p>
```
Just for dev purposes. I was really confused when reviewing about empty steps.

EdoStorm96: Or alternatively, remove unimplemented question types from dev data generation, as I suggested in another comment.

backend/form/models.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class FormConfig(models.Model):
+    name = models.CharField(max_length=200)
+    version = models.CharField(max_length=100)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+
+class Form(models.Model):
```
EdoStorm96: Should we call it `MRForm` and `MRFormConfig` to avoid confusion with Django forms?

backend/form/models.py
```diff
@@ -0,0 +1,231 @@
...
+# Questions
+class BaseQuestion(models.Model):
+    text = models.CharField(max_length=200)
+    step = models.ForeignKey(Step, on_delete=models.CASCADE)
+    description = models.TextField(
+        blank=True,
+        help_text="Context about the question and its purpose, which will be shown to the user.",
+    )
+    required = models.BooleanField(default=False)
+
```
EdoStorm96, line 78: Should questions also have an order within a step? I think so ... I'd like to also have a similar constraint as with steps, where it checks that questions of the same step all have a unique order.

backend/form/models.py
EdoStorm96: Sorry for not providing a correct QS way to do this haha ... But my head is already spinning from this review ;p

EdoStorm96 (commented):
Alright! Amazing work! I am very impressed by your foresight and accuracy. I think it is all looking really good and clean. I had a careful look at everything and it all looks really good. I could not find any buggy or messy things.

I must admit that I don't feel like I fully grasp everything that is going on ... So I'd really prefer if you let someone else have a look at this too. I do approve-ish ... But it'd be nice to get some more input on this, so just a comment right now.

I do have some comments/suggestions/questions but mostly nothing I feel super strongly about.

I do think we need to think a bit more about how we implement order. Not only for steps within the form, but also for questions within a step. This gets a bit complicated, but I have some ideas for the logic in my comments.

The only larger thing I'd like to discuss a bit, is the roles of `Form` and `FormConfig`. Cause now I feel like we could just merge these into one model.

Anyway, awesome job!

frontend/components/shared/MRForm.vue
```diff
@@ -37,14 +57,10 @@ const formStepperConfig = computed<FormStepperConfig>(() => {
...
-
-function getAllSteps(): FormStep[] {
-    const allSteps: FormStep[] = [];
+function getAllSteps(): CombinedStep[] {
+    const allSteps: CombinedStep[] = [];
 
-    props.formConfig.steps.forEach((step) => {
+    steps.value.forEach((step) => {
         // Add main step.
         allSteps.push(step);
```
RoordinkFrank: This line has an warning.
image
Perhaps due to the fact that there is mockdata involved?

frontend/components/shared/MRForm.vue
```diff
@@ -1,33 +1,53 @@
...
+    form: QueriedForm;
 }
 
 const props = defineProps<Props>();
 
+const steps = computed<ParentStep[]>(() =>
+    props.form.steps.map((step) => ({
+        ...step,
+        substeps:
+            step.substeps?.map((substep) => ({
```
RoordinkFrank: image
Unlike the red undescore suggests this is merely a warning.

frontend/components/shared/MRForm.vue
```diff
@@ -1,33 +1,53 @@
...
 import { BSButton } from "cdh-vue-lib";
-import type { FormConfig, FormStep } from "../form/types";
 import FormStepper, { type FormStepperConfig } from "./FormStepper.vue";
+import { type GetFormQuery } from "~/generated/gql/graphql";
+
+type QueriedForm = NonNullable<GetFormQuery["form"]>;
+type ParentStep = NonNullable<QueriedForm["steps"][number]>;
+type Substep = NonNullable<ParentStep["substeps"][number]>;
+
+export type CombinedStep = ParentStep | Substep;
```
RoordinkFrank: the naming of CombinedStep does not immediately clear what it does. In reality what you mean here is a step no matter what kind of step. Abstraction would be great here.

frontend/components/shared/MRForm.vue
```diff
@@ -1,33 +1,53 @@
 <script lang="ts" setup>
 import { ref, computed } from "vue";
 import { BSButton } from "cdh-vue-lib";
-import type { FormConfig, FormStep } from "../form/types";
 import FormStepper, { type FormStepperConfig } from "./FormStepper.vue";
+import { type GetFormQuery } from "~/generated/gql/graphql";
+
+type QueriedForm = NonNullable<GetFormQuery["form"]>;
```
RoordinkFrank: QueriedForm reads a bit weird. It can be interpreted that the type is a form as well it is called form. When it is still a query according to the "query GetForm" in index. You are making it in a form but at this point it is still data if I interpret this correctly.
I want to say keep the naming as [name]FormQuery. Though in this case that would be formformquery so a bit weird as well.

RoordinkFrank (changes requested):

frontend/components/shared/MRForm.vue
```diff
@@ -1,33 +1,53 @@
...
             slug: step.slug,
-            labelNl: step.stepNameNL,
-            labelEn: step.stepNameEN,
+            labelNl: step.nameNl ?? "",
+            labelEn: step.nameEn ?? "",
             completed: false,
             active: step.slug === selectedStepSlug,
             disabled: false,
             children:
                 step.substeps?.map((substep) => ({
```
RoordinkFrank: Same warning as above, 
image

backend/form/models.py
```diff
@@ -0,0 +1,231 @@
...
+        max_length=200,
+        unique=True,
+        help_text="Used in the URL.",
+    )
+
+    form = models.ForeignKey(
+        Form, null=True, blank=True, on_delete=models.CASCADE, related_name="steps"
+    )
+    form_order = models.PositiveIntegerField(
+        null=True, blank=True, help_text="The order of this step within the form."
```
RoordinkFrank: image
I do like the helpful comments but these comments seem meant for us. The help_text field seems be meant for users.

frontend/components/shared/FormSideBar.vue
XanderVertegaal: It did for me as well! This is some pretty advanced GraphQL stuff that I wasn't aware about either 😄

backend/form/models.py
XanderVertegaal: Not a bad idea. I don't think we'll be using Django `Form`s much, but it's good to avoid overlap. 👌

backend/form/models.py
XanderVertegaal: Yes, but our users are never going to see these texts, so it's fine 😉 These help texts show up in Django `forms` and in the admin environment. We won't be using the former and the latter is for our eyes only, so it doesn't hurt to use `help_text` for some additional documentation here.

backend/form/models.py
XanderVertegaal: Don't worry! I think the first two points are already included in this check, but the latter two are good additions.

backend/main/management/commands/create_dev_data.py
XanderVertegaal: Aye-aye!

backend/form/models.py
XanderVertegaal: Currently, I can't really think of any reasons not to cut out the middle man, but I'd like to discuss this first. There is an issue about it now: https://github.com/CentreForDigitalHumanities/myresearch/issues/55

backend/form/models.py
XanderVertegaal: I've added a field `step_order` and constraint, but it only checks if questions of the same type (text, date, etc.) have a unique value for this field within a step. The constraint only stops two TextQuestions from having the same order, but not a TextQuestion and a DateQuestion (since those are two different models).

We can implement a clean method for this in the future. I'll create an issue for this.

backend/form/types/QuestionType.py
XanderVertegaal: Yes, a GraphQL Interface basically defines fields that are shared by multiple Types. Every Type implementing an Interface will have that Interface's fields as well. It's very similar to extending an abstract class in Django models. I guess the "id" field is superfluous in his case.

frontend/components/shared/FormStep.vue
XanderVertegaal: I'm implementing this solution (emoji included 😉).

backend/main/management/commands/create_dev_data.py
XanderVertegaal: I'm implementing your placeholder in the frontend.

frontend/components/shared/MRForm.vue
XanderVertegaal: You're right that it's still data at this point (and not a form object yet). It's not a query, however. Here I'm extracting the form from the query result (which potentially has error messages etc.). I think calling the type `...Form` makes sense here.

frontend/components/shared/MRForm.vue
XanderVertegaal: It's a yellow squiggly on my end :) But yes, nullish coalescing (`??`) is not needed anymore. I've removed it!

frontend/components/shared/MRForm.vue
XanderVertegaal: No, `step.substeps` used to be a nullable field, but it is now non-nullable. Good find!

frontend/components/shared/MRForm.vue
XanderVertegaal: I've renamed it to `Step`.

backend/api/graphql/schema.py
tijmenbaarda: Perhaps add this as a comment in the code?

backend/form/models.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+class StepInfo(models.Model):
+    """
+    More information and FAQ for a specific Step within a Form.
+    """
+
+    step = models.OneToOneField(
+        Step, null=True, blank=True, on_delete=models.CASCADE, related_name="info"
+    )
```
tijmenbaarda: What is the reason to include a separate model here with a one-to-one relationship?

backend/form/models.py
```diff
@@ -0,0 +1,257 @@
...
+    required = models.BooleanField(default=False)
+
+    class Meta:
+        abstract = True
+        constraints = [
+            models.UniqueConstraint(
+                fields=["step_id", "step_order"],
+                name="%(app_label)s_%(class)s_unique_step_order",
+            )
+        ]
```
tijmenbaarda: All question types get separate database tables, so this will not enforce that there be no questions with duplicate orders if they are of different types.

A way to solve this would be to use non-abstract base models, or to use one model for all question types.

(I am also concerned that by enforcing this on the database level it will get difficult to change the order without running into conflicts: swapping two questions, for example, will only be possible within a transaction, but that is not possible if done from within the Django admin pages.)

backend/form/models.py
```diff
@@ -0,0 +1,257 @@
...
+    )
+
+
+class FileUploadQuestionAnswer(models.Model):
+    question = models.ForeignKey(
+        FileUploadQuestion, on_delete=models.CASCADE, related_name="answers"
+    )
+    answer = models.ForeignKey(
+        FileUploadAnswer, on_delete=models.CASCADE, related_name="question_answers"
+    )
```
tijmenbaarda: General remark: would it be possible to use non-abstract base classes? That would eliminate the need to define different models for the question-answer combinations, because this seems to become a little bit too complicated...

I remember trying to use non-abstract base models in the past and I was disappointed because it was unstraightforward to get to the specific model from the generic model - this was only possible with an extension. But this would be so useful for our case that I think we could explore it further...

backend/form/models.py
```diff
@@ -0,0 +1,257 @@
...
+
+class FileUploadQuestion(BaseQuestion):
+    size_limit = models.PositiveIntegerField()
+
+
+# Answers
+class BaseAnswer(models.Model):
+    # This means you can access true_false_answers with <user>.truefalseanswers etc.
+    user = models.ForeignKey(
+        user_model, on_delete=models.CASCADE, related_name="%(class)ss"
```
tijmenbaarda: Maybe not `CASCADE` but `PROTECT`? Although in practice users should probably not be deleted. Great idea not to hardcode the user model here!

I suppose a relation to an `Application` object is to be added in the future?

backend/form/types/MRFormType.py
```diff
@@ -0,0 +1,47 @@
...
+from form.models import MRForm, MRFormConfig, Step
+
+
+class MRFormType(DjangoObjectType):
+
+    class Meta:
+        model = MRForm
+        fields = [
+            "id",
+            "name_nl",
```
tijmenbaarda, line 16: Is there no way to use just `"name"` and let the language be chosen automatically based on the language setting in the Django session storage? (I would be happy to look into this, but I can imagine that we have tried this before and failed :)

backend/main/management/commands/create_dev_data.py
```diff
@@ -35,6 +106,181 @@ def handle(self, *args, **options):
...
+        )
+
+        print("Done!")
+
+        return form
+
+    def _generate_steps(self, options, form: MRForm) -> None:
+        number_of_steps = self.faker.random_int(MIN_STEPS_PER_FORM, MAX_STEPS_PER_FORM)
+
+        for step_index in tqdm(range(number_of_steps), desc="Generating steps..."):
```
tijmenbaarda: Good idea to use `tqdm` :)

tijmenbaarda (approved):
Wonderful! I am amazed by the low amount of code that was needed in the frontend to make this work.

I mainly have some remarks in the model definitions. We already discussed most of this, but I am having some second thoughts now that I see the complexity that the "small models" come with... My main suggestion is to explore the use of non-abstract base classes ([multi-table inheritance](https://docs.djangoproject.com/en/5.2/topics/db/models/#multi-table-inheritance)) - I know that this is not straightforward but it would potentially help a lot for our use case to reduce complexity.

backend/main/management/commands/create_dev_data.py
```diff
@@ -35,6 +106,181 @@ def handle(self, *args, **options):
         with transaction.atomic():
             self._create_test_users(options)
 
+            form = self._generate_form(options)
+            self._generate_steps(options, form)
+            self._generate_questions(options, form)
```
miggol: I don't think the form generation and user import should be in the same transaction.

Actually I think it would be useful to have them in separate commands, allowing the regeneration of some forms without overwriting changes you may have made to your users.

This request isn't urgent, we can always improve on this down the road.

backend/form/models.py
```diff
@@ -0,0 +1,257 @@
...
+    )
+
+    parent = models.ForeignKey(
+        "self", null=True, blank=True, on_delete=models.CASCADE, related_name="substeps"
+    )
+    parent_order = models.PositiveIntegerField(
+        null=True,
+        blank=True,
+        help_text="The order of this step within its parent step.",
+    )
```
miggol: Perhaps I'm oblivious to some frontend constraints, but I'm feeling some redundancy in these attributes.

1. Can't we just have a single order, and then have parent and order be unique together?

2. The above would require that every form would have a "root step" that is the top parent, otherwise the orders of orphans (with parent=null) would clash. This step could provide the name and description for the entire form, building Edo's case for the redundancy of the Form object.

Let me know what you think. I like this idea because it makes the form self-similar at every "zoom level", should we ever only want to render or query a subform.

3. As for FormConfig, we could either have every step link to it as we now do with the Form object. But we could also give the FormConfig a "root_step" attribute which points to the entrypoint from which the form is built, by recursively visiting all of its children.

It's early days, but I think the latter might be useful if we want a separate FormConfig for the subset of a form, e.g. the repeating "Participant Group" section, that we might edit sometimes but otherwise keep the same.

backend/form/models.py
miggol: I would like to second the non-abstract base model idea, I think these question types are an ideal candidate for that.

For orders in Django-admin you can write "Move up" / "Move down" admin actions and enable them on every model that has an order. 

I've been waiting for an excuse to write an admin action for a while now, this kind of thing is exactly what they're made for: https://docs.djangoproject.com/en/5.2/ref/contrib/admin/actions/

frontend/components/shared/MRForm.vue
```diff
@@ -1,60 +1,74 @@
...
 const props = defineProps<Props>();
 
+const steps = computed<ParentStep[]>(() =>
+    props.form.steps.map((step) => ({
+        ...step,
+        substeps: step.substeps.map((substep) => ({
+            ...substep,
+        })),
+    })),
+);
```
miggol: I like it when JS gets all functional like this. Though a recursive implementation would be more generic once we increase step depth > 1.

Though I can see why this is better for now. No need to worry about levels to compute the indentation and whatnot.

miggol (changes requested):
Thanks for bringing all this to the table, Xander. I learned a lot reviewing this! Sorry it took a while

My main remarks are with the model structure on the backend side. I think these are worth addressing now rather than soldiering on

frontend/components/shared/MRForm.vue
XanderVertegaal: I played around with having 'infinite' depth, but it proved to be non-trivial and I wasn't sure we needed the added complexity. We can discuss this on Wednesday, maybe!

backend/form/models.py
XanderVertegaal: `PROTECT` is a good idea! And yes: there should indeed be some link to an application. That could go via the top-level Form/FormConfig object and/or the QuestionAnswer model (if we decide to keep that). I'll leave that for a future PR, though!

backend/form/models.py
XanderVertegaal: Thanks for the suggestion guys! I'll definitely look into this :) If we use non-abstract base models we might also be able to leverage [Django's built-in `order_with_respect_to`](https://docs.djangoproject.com/en/5.2/ref/models/options/#order-with-respect-to).

backend/form/models.py
XanderVertegaal: It felt natural at the time in the light of separation of concerns, but I agree that this does not warrant a separation of models. I'll try to implement `info` as a field on `Step`!

backend/form/models.py
XanderVertegaal: Great suggestion, I like this idea too. 👌 I'll try to merge Forms and Steps (perhaps into a single `Form` model). The added benefit (or risk) is that we can then start nesting Forms ad infinitum ;)

backend/form/models.py
XanderVertegaal: I'll look into this!

backend/main/management/commands/create_dev_data.py
XanderVertegaal: No, you're right. They should be separated. I'll remove the user import from this command (and perhaps create a super-simple second command for this).

backend/form/types/MRFormType.py
XanderVertegaal: That sounds very fancy. I have never seen an implementation that does it, but it does not sound impossible and I cannot say I have considered this. Feel free to try this after this is merged.

XanderVertegaal:
Thank you all for your feedback! <3

Based on your suggestions, the data model and implementation could be simplified drastically.

The relevant changes compared to the first review are the following.

- MRForm and MRFormConfig have been merged into one model (@EdoStorm96).
- Steps have been removed. There is now only one MRForm model with a field 'parent' that refers to itself. Consequently, we now have one top-level form with nested subforms, which allows for nesting up to an arbitrary depth (@miggol). 
  - NB: the Stepper component only supports up to five, and the dev data creation script only generates up to two, so we stay sane :P
- MRForm now uses [Django's built-in 'order_with_respect_to'](https://docs.djangoproject.com/en/5.2/ref/models/options/#order-with-respect-to) to generate an `_order` field, which governs the order of subforms within a parent form.
- The individual question-related models now inherit the non-abstract BaseQuestion model (@tijmenbaarda, @miggol). This also implements `order_with_respect_to` to ensure the order of questions within a (sub)form.
- "StepInfo" is no longer needed as a separate model (@tijmenbaarda). Additional info displayed in the sidebar in the frontend is now implemented in the form of simple fields.
- Answer models and QuestionAnswer models have been removed (for now). Their implementation warrants further discussion and a separate PR. The API did not expose these models anyway, so not much is lost here.

Open questions (will be converted into issues once merged):
- The field 'version' used to be present on MRFormConfig, but since that is now merged with MRForm, it is now present on every form/subform. This is technically speaking redundant. Do we mind?
- The GraphQL query retrieving the form is massive with lots of duplication. We can use GraphQL Fragments to simplify/DRY this, but this is not trivial.
- I have removed _create_test_users from the transaction that also holds the form generation methods, but I have kept it as part of the create_dev_data.py script for now. Moving it to a separate file breaks the _check_all_models_implemented check. We should discuss how we want to solve this.
- Automatic selection of translatable fields (`name_en`/`name_nl`) based on Session locale.

XanderVertegaal:
(Small addition, since this only took a few minutes: components for TrueFalseQuestion, FileUploadQuestion and NumberQuestion. Now all question types have a corresponding frontend control.)

EdoStorm96 (commented):
Ok, I've had a quick look at this and it is looking really great I think. Some very nice improvements, which makes this a lot more intuitive.  order_with_respect_to is an awesome find!

However, I do have one new gripe ... Because now that steps have been removed, I do kindoff feel like the toplevel Form model should somehow be distuinguished ... It is now a bit strange that this gets treated as a normal form, even though it does not contain any questions. This is currently noticeable when visiting the mock procreg. You start out on a page with no questions and the stepper also has one funky step with no bubble. There are workarounds for this of course ... But I think the toplevel 'form' should be clearly recognizable.

So what I am trying to say ... Now that steps have been removed ... I think we might need MRFormConfig again ... 😅 It is a bit weird to have a toplevel object with really a different role, that has all the same attributes as its children ... This would also solve the problem of each Form having a version.

I am not 100% sure if then all form should be related to the config or only the toplevel ones. I suppose all forms could be related to the config... But there would need to be a good way to find the first form via the config.

Does anyone else feel this way?

Anyway, really great work again :)

XanderVertegaal:
Thanks @EdoStorm96 for taking the time to have another look and note your thoughts! ❤️

The current approach does indeed have its downsides, as you notice.
- the top-level form uses the same model as its subforms (which can be confusing);
- the `version` field now lives on all (sub)forms, which is redundant.
- the top-level form object has a `questions` field, which is only relevant for subforms

At the same time, there are workarounds/benefits.
- Top-level forms are always parent-less, so `MRForm.objects.filter(parent__isnull=True)` should give you all top-level forms.
- You can easily reuse subforms as top-level forms and vice versa, so composing new forms based on existing elements is easy.
- There is a certain elegance/simplicity in using one model for all levels. (I immediately concede that this is a matter of taste/style.)

In any case, I will bring this up during the meeting tomorrow! :) I am curious to hear what the others think.

frontend/components/layout/NavBar.vue
```diff
@@ -1,8 +1,22 @@
...
+    query GetFirstSlug {
+        form {
+            id
+            slug
+        }
+    }
+`);
+
+const { result } = useQuery<GetFirstSlugQuery>(GET_FIRST_SLUG);
+const slug = computed<string | null>(() => result.value?.form?.slug ?? null);
```
EdoStorm96: To avoid visiting top level form from the navbar:
```suggestion
const GET_FIRST_SLUG = graphql(`
    query GetFirstSlug {
        form {
            id
            subforms {
                id
                slug
            }
        }
    }
`);

const { result } = useQuery<GetFirstSlugQuery>(GET_FIRST_SLUG);
const slug = computed<string | null>(() => result.value?.form?.subforms?.[0]?.slug ?? null);
```

frontend/components/shared/FormStepper.vue
```diff
@@ -43,12 +43,12 @@ function stepperItemClasses(step: FormStep): string {
     <div class="stepper-container">
         <div class="stepper">
             <p class="mb-4">
-                {{ useTranslateableAttribute(props.stepConfig, "title") }}
+                {{ useTranslateableAttribute(props.stepperConfig, "title") }}
             </p>
```
EdoStorm96: I'd say get rid of this for now.

frontend/pages/procreg/[slug].vue
```diff
@@ -0,0 +1,154 @@
+<script lang="ts" setup>
+import { SharedFormWrapper } from "#components";
+import { useQuery } from "@vue/apollo-composable";
+import { graphql } from "~/generated/gql";
+import type { GetFormQuery } from "~/generated/gql/graphql";
+
+const GET_FORM = graphql(`
+    query GetForm {
+        form {
```
Comment (EdoStorm96) [inline frontend/pages/procreg/[slug].vue:9]: I guess the toplevel form does not really need all the info of a normal for right? Maybe only id and subforms? Or maybe version?

frontend/components/shared/FormWrapper.vue
```diff
@@ -0,0 +1,146 @@
...
+
+function getAllForms(form: CombinedForm) {
+    const collectedForms = [form];
+    if ("subforms" in form) {
+        form.subforms.forEach((subform) => {
+            collectedForms.push(...getAllForms(subform));
+        });
+    }
+    return collectedForms;
+}
```
EdoStorm96: Maybe this can become something like this:
```suggestion
function getAllForms(form: CombinedForm) {
    const collectedForms: CombinedForm[] = [];
    if ("subforms" in form && Array.isArray(form.subforms)) {
        form.subforms.forEach((subform) => {
            collectedForms.push(subform, ...getAllForms(subform));
        });
    }
    return collectedForms;
}
```
So that the top level form gets excluded from the normal forms again :)

frontend/components/shared/FormWrapper.vue
```diff
@@ -0,0 +1,146 @@
...
+
+const formStepperConfig = computed<FormStepperConfig>(() => {
+    const selectedFormSlug = props.currentFormSlug;
+
+    const topLevelForm = props.form;
+
+    return {
+        titleNl: topLevelForm.nameNl ?? "",
+        titleEn: topLevelForm.nameEn ?? "",
+        steps: topLevelForm.subforms.map((form) => ({
```
EdoStorm96: Maybe rename steps to subforms?

frontend/components/shared/FormWrapper.vue
```diff
@@ -0,0 +1,146 @@
...
+            slug: form.slug,
+            labelNl: form.nameNl ?? "",
+            labelEn: form.nameEn ?? "",
+            completed: false,
+            active: form.slug === selectedFormSlug,
+            disabled: false,
+            children: form.subforms.map((subform) => ({
```
EdoStorm96: And children to subforms as well. Mutatis mutandis ;p

EdoStorm96 (commented):
Ok, I see your point and I do also see the appeal of this design. Then I'd like to see the rest of this branch get updated a tiny bit to reflect the new design. I've made some final suggestions (thanks copilot ...) to make the frontend work like I think it should now. I think incorporating this will make it nicer for the other reviewers as well.

Oh and as a general note, I think we should try to be consistent about the naming of subforms, so lets try to avoid steps and children to avoid confusion.

All nitpicks made out of love and respect <3

Ps. if you know of better ways to accomplish what my suggestions accomplish, I am curious to learn!

XanderVertegaal:
During today's meeting, we decided to reinstate a separate top-level model (FormConfig vel sim.) for now! If we don't like it or require more flexibility, we can always change it back. I think this strikes a good balance between elegance and compositionality.

XanderVertegaal:
Alright, final changes before merging.

There is now a single top-level `MRForm` model with infinitely nestable `Step` objects. There is no more `version` field, as discussed, but the fields `created_at` and `updated_at` have been kept to make sure we can track legacy forms.

A `Step` can have a form (referring to the top-level `MRForm`) or a parent `Step` object, but not both.

(I apologize for going back and forth when it comes to the naming of these models! 🙈)

PR #50, opened 2025-08-26
Feature/links to form steps

EdoStorm96:
So I thought it would be nice if the separate questions have unique urls, and that the stepper is clickable. So that is this PR. It was very easy to get this working. One thing to keep in mind is that the folder `proc-reg` in pages has been renamed to `procreg`. The `-` interferes with the syntax used in NavBar.vue, line 49.

Otherwise, this was a easy as I´d hoped :)

(Oh, and my typescript might be terrible ... So do pay special attention to that.


frontend/components/layout/NavBar.vue
```diff
@@ -43,7 +46,7 @@ import { BSIcon } from "cdh-vue-lib";
                     </li>
                     <li>
                         <NuxtLink
-                            to="/proc-reg"
+                            :to="{ name: 'procreg-step', params: { step: firstProcRegSlug } }"
```
XanderVertegaal: Nice! I never knew you could pass in an object with `name` and `params` in this way. 

(I would have used `:to="procreg-step/${firstProcRegSlug}"` or something like that, but I like this more!)

frontend/components/layout/NavBar.vue
```diff
@@ -1,5 +1,8 @@
 <script lang="ts" setup>
 import { BSIcon } from "cdh-vue-lib";
+import { mockVwr } from "~/components/form/procRegMockConfig";
+
+const firstProcRegSlug = mockVwr.steps[0].slug;
```
XanderVertegaal, line 5: Interesting that TS does not complain about this. Technically, `mockVwr.steps[0]` could be undefined if `mockVwr.steps` is empty. This could lead to runtime errors.

(This is me learning something I didn't know about TS, not something that needs to be changed :))

XanderVertegaal: Update: there is apparently a TS setting for this, although it is called 'pedantic' in [this StackExchange issue](https://stackoverflow.com/questions/50647399/typescript-accessing-an-array-element-does-not-account-for-the-possibility-of-u).

frontend/pages/gql_tst.vue
```diff
@@ -8,26 +8,26 @@ import { type GetAllUsersQuery } from "~/generated/gql/graphql";
...
 </script>
 
 <template>
-  <div>
-    <p v-for="user in usersResult?.users ?? []">
-      {{ user.username }} {{ user.email }}
-    </p>
-  </div>
+    <div>
+        <p v-for="user in usersResult?.users ?? []">
```
XanderVertegaal: I'm getting linting errors on this line: `Elements in iteration expect to have 'v-bind:key' directives.eslintvue/require-v-for-key`. This is not due to anything you did, but if you get errors here as well, you might as well fix it by committing:


```suggestion
        <p v-for="user in usersResult?.users ?? []" :key="user.id">
```

frontend/pages/procreg/[step].vue
```diff
@@ -11,7 +17,10 @@ const formConfig = mockVwr;
             <h1>{{ $t("Processing Registry") }}</h1>
         </div>
         <div class="uu-container">
-            <SharedMRForm :form-config="formConfig" />
+            <SharedMRForm
+                :form-config="formConfig"
+                :currentStepSlug="stepSlug(route.params.step)"
```
Comment (XanderVertegaal) [inline frontend/pages/procreg/[step].vue]: I'm getting a linting warning here: `Attribute ':currentStepSlug' must be hyphenated.eslintvue/attribute-hyphenation`


```suggestion
                :current-step-slug="stepSlug(route.params.step)"
```

Your code still works, but it is conventional to use kebab case, even in Vue templates. (Vue will automatically convert it to camel case).

frontend/components/shared/MRForm.vue
```diff
@@ -38,7 +39,10 @@ const formStepperConfig = computed<FormStepperConfig>(() => {
 });
 
 const selectedStep = ref<FormStep | null>(
-    props.formConfig.steps.length > 0 ? props.formConfig.steps[0] : null,
+    props.formConfig.steps.length > 0
```
XanderVertegaal: Logically, this works great, but nesting nullish coalescing `... ?? ...` inside of a ternary expression `... ? ... : ...` is not easy to read (in my opinion)

You may want to outfactor this to a special function, but this is not required!

```ts
function defaultStep(): FormStep | null {
    const allSteps = getAllSteps();
    if (allSteps.length <= 0) {
        return null;
    }
    return allSteps.find(({ slug }) => slug === props.currentStepSlug) ?? null;
}

const selectedStep = ref<FormStep | null>(defaultStep());
```

XanderVertegaal (approved):
Nice, this seems to work really well! Glad to hear this did not turn out to be very difficult in the end. :) I have added a few stylistic comments, but definitely nothing that should stop you from merging (as far as I'm concerned).

PR #52, opened 2025-08-27
Feature coverimage

### LINKED ISSUE #51: add coverimage to frontpage
source code is on /ethics but that's in python

RoordinkFrank:
subbranch of frontpage branch. Only the added Seasonsal Cover commit is relevant.


XanderVertegaal:
I've changed the base of this PR, so it is now based on `feat/frontpage` instead of `develop`. That makes it a lot cleaner/smaller :)

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,57 @@
...
+  },
+  winter: {
+    image: winterURL,
+    author: "Simona Evstatieva",
+    classes: "align-items-middle",
+  },
+};
+//only considers Dutch seasons, not Australian
+function getSeasonalCoverImage() {
+  let month = new Date().getMonth() + 1; //const month goes from 1 to 12
```
XanderVertegaal, line 31: I'd use `const` instead of `let` here. In general, you should try to use `const` whenever you can and use `let` if you really need a mutable variable.

I'd also remove the comment at the end of the line. It's clear from the rest of the code what you're doing here.

```suggestion
  const month = new Date().getMonth() + 1;
```

XanderVertegaal:
```suggestion
  const month = new Date().getMonth() + 1;
```

General JS/TS tip: try to use `const` instead of `let` whenever you can.

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,57 @@
...
+  if ([3, 4, 5].indexOf(month) >= 0) {
```
XanderVertegaal:
```suggestion
  if ([3, 4, 5].includes(month)) {
```

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,57 @@
...
+import springURL from "@/assets/images/covers/coverimage-spring.jpg";
+import summerURL from "@/assets/images/covers/coverimage-summer.jpg";
+import autumnURL from "@/assets/images/covers/coverimage-autumn.jpg";
+import winterURL from "@/assets/images/covers/coverimage-winter.jpg";
+const seasons = {
+  spring: {
+    image: springURL,
+    author: "Dick Boetekees",
+    // seasons.classes controls how the image is fitted inside the container; every image has its own 'best' fit
+    classes: "align-items-end",
```
XanderVertegaal: I don't see a difference if I disable these classes. Are you sure they are needed?

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,57 @@
+<script setup lang="ts">
+import springURL from "@/assets/images/covers/coverimage-spring.jpg";
+import summerURL from "@/assets/images/covers/coverimage-summer.jpg";
+import autumnURL from "@/assets/images/covers/coverimage-autumn.jpg";
+import winterURL from "@/assets/images/covers/coverimage-winter.jpg";
+const seasons = {
```
XanderVertegaal, line 7: Add a space after the last import for clarity.

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,57 @@
...
+    author: "Ivar Pel",
+    classes: "align-items-middle",
+  },
+  winter: {
+    image: winterURL,
+    author: "Simona Evstatieva",
+    classes: "align-items-middle",
+  },
+};
+//only considers Dutch seasons, not Australian
```
XanderVertegaal: This comment made me laugh 😂, but it's a bit confusing, so perhaps it's best to remove it 😉

XanderVertegaal (commented):
Nice work! The seasonal images look great :) It would be great if you could check `classes` one more time, though.

frontend/components/SeasonalCover.vue
RoordinkFrank: I am not 100% sure. There is no difference for our current placement of the cover. This css part  are originally from /Ethics and I did not dare to change them. Perhaps they will matter if our cover location changes position on the website.

frontend/components/SeasonalCover.vue
RoordinkFrank: seems slightly cleaner indeed

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,57 @@
...
+    image: winterURL,
+    author: "Simona Evstatieva",
+    classes: "align-items-middle",
+  },
+};
+function getSeasonalCoverImage() {
+  const month = new Date().getMonth() + 1; //months go from 1 to 12.
+  if ([3, 4, 5].includes(month)) {
+    return seasons["spring"];
+  } else if ([6, 7, 8].indexOf(month) >= 0) {
```
XanderVertegaal: These `indexOf`s can also be replaced by an `.includes()` (to keep things simple and consistent).

frontend/components/SeasonalCover.vue
XanderVertegaal: In that case I'd remove them to keep things simple. We can always re-add them if we need to :) If you want, you can add a comment referring to Ethics so we know where to look if we run into positioning issues.

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,54 @@
+<script setup lang="ts">
+import springURL from "@/assets/images/covers/coverimage-spring.jpg";
+import summerURL from "@/assets/images/covers/coverimage-summer.jpg";
+import autumnURL from "@/assets/images/covers/coverimage-autumn.jpg";
+import winterURL from "@/assets/images/covers/coverimage-winter.jpg";
+
+const seasons = {
+  //originally the images that come from /Ethics had css classes as well,
+  // but they seemed no longer needed.
```
XanderVertegaal, line 9:
```suggestion
  // but they seem to be no longer needed.
```

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,54 @@
...
+    return seasons["winter"];
+  }
+}
+const seasonCover = getSeasonalCoverImage();
+</script>
+
+<template>
+  <div class="uu-cover h-400">
+    <img
+      v-bind:src="seasonCover.image"
```
XanderVertegaal:
```suggestion
      :src="seasonCover.image"
```

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,54 @@
...
+      class="uu-cover-image w-100"
+      alt="cover"
+    />
+    <div class="uu-cover-copyright uu-cover-copyright-right">
+      Cover by {{ seasonCover.author }}
```
XanderVertegaal:
```suggestion
      {{ $t("Cover by") }} {{ seasonCover.author }}
```

frontend/components/SeasonalCover.vue
```diff
@@ -0,0 +1,54 @@
+<script setup lang="ts">
+import springURL from "@/assets/images/covers/coverimage-spring.jpg";
+import summerURL from "@/assets/images/covers/coverimage-summer.jpg";
+import autumnURL from "@/assets/images/covers/coverimage-autumn.jpg";
+import winterURL from "@/assets/images/covers/coverimage-winter.jpg";
+
+const seasons = {
+  //originally the images that come from /Ethics had css classes as well,
```
XanderVertegaal:
```suggestion
  // Originally the images that come from Ethics had CSS classes as well,
```

XanderVertegaal (approved):
Well done! A few more nitpicks, but otherwise ready to merge! :D

PR #58, opened 2025-09-03
feat: add mutation to gql_tst

### LINKED ISSUE #57: Simple gql mutation
So now that we have a proof of concept for getting data from the backend to the frontend (AKA a query), it'd also be nice to get data from the frontend to the backend (AKA a Mutation)

EdoStorm96:
fixes #57.

So this is a really simple proof of concept for getting data from the frontend to the backend using a form. On http://127.0.0.1:3000/gql_tst/, you can now not only see all users, but also create new ones. Pretty simple stuff, but it seems to work.

I brute forced my way though again, so I'm open to any feedback :)


backend/api/graphql/schema.py
```diff
@@ -1,10 +1,29 @@
...
+    create_user = CreateUser.Field()
+    update_user = UpdateUser.Field()
+    delete_user = DeleteUser.Field()
+
+class Mutation(ObjectType):
+
+    create_user = CreateUser.Field()
+    update_user = UpdateUser.Field()
+    delete_user = DeleteUser.Field()
+
```
XanderVertegaal, line 20: I think you've accidentally doubled the mutation code here ;)

backend/main/mutations/UserMutations.py
```diff
@@ -0,0 +1,55 @@
+import graphene
+
+from main.types.UserType import UserType
+from django.contrib.auth import get_user_model
+    
+class CreateUser(graphene.Mutation):
```
XanderVertegaal: For consistency's sake, I recommend you import from graphene what you need (i.e. `from graphene import ...`). It also saves you from prepending everything with `graphene.`.

backend/main/mutations/UserMutations.py
```diff
@@ -0,0 +1,55 @@
...
+    class Arguments:
+        username = graphene.String(required=True)
+        email = graphene.String(required=True)
+
+    user = graphene.Field(UserType)
```
XanderVertegaal: Mutation return fields (in this case `user`) are usually put before the `Arguments` subclass (but this is just a conventional thing).

frontend/pages/gql_tst.vue
```diff
@@ -17,17 +18,92 @@ const GET_USERS = graphql(`
...
+      />
+      <p>
+          E-mail:
+      </p>
+      <input
+          type="text"
+          class="form-control"
+          v-model="email"
+      />
+      </div>
```
XanderVertegaal, line 116: `v-model` attributes should be the first mentioned.

XanderVertegaal, line 116: I'm getting an error here: `Elements in iteration expect to have 'v-bind:key'`. Not a biggie since this is for testing anyway, but it's good to keep in mind.

backend/main/mutations/UserMutations.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def mutate(self, info, username, email):
+        user = get_user_model()(username=username, email=email)
```
XanderVertegaal, line 43: This is very confusing, but the way the `Graphene` docs describe how mutations should be called is not recommended. (The `Graphene-Django` docs are more explicit, see also this discussion if you're interested: https://github.com/graphql-python/graphene/issues/810.)

The heart of the matter is that `mutate` is not really a regular method, so if you inspect `self`, it will return as `None`. This means you can't use other class methods that you might have defined within the `CreateUser` class (which is silly). The Graphene-Django docs recommend marking the function as a `@classmethod`. The `cls` parameter then gives you access to other methods defined in this class. The argument `root` can be ignored (it refers to the 'parent' object, which is relevant for Types that are fields on other Types in the schema, but Mutations are top-level entities, so their root/parent is always `None`).

The suggested signature for this method (and mutatis mutandis also for the other mutations in this file) is:

```suggestion
from graphene import ResolveInfo

@classmethod
def mutate(cls, root: None, info: ResolveInfo, username: str, email: str):
```

backend/main/mutations/UserMutations.py
XanderVertegaal: You may also want to include something like an `error` field to any mutation to notify the client of any errors/warnings that happened during execution of the mutation.

backend/main/mutations/UserMutations.py
```diff
@@ -0,0 +1,55 @@
...
+class DeleteUser(graphene.Mutation):
+    class Arguments:
+        id = graphene.ID(required=True)
+
+    success = graphene.Boolean()
+
+    def mutate(self, info, id):
+        try:
+            user = get_user_model().objects.get(pk=id)
+        except user.DoesNotExist:
```
XanderVertegaal: For consistency:

```suggestion
        except get_user_model().DoesNotExist:
```

backend/main/mutations/UserMutations.py
```diff
@@ -0,0 +1,55 @@
...
+        username = graphene.String(required=True)
+        email = graphene.String(required=True)
+
+    user = graphene.Field(UserType)
+
+    def mutate(self, info, id, username=None, email=None):
+        try:
+            user = get_user_model().objects.get(pk=id)
+        except get_user_model().DoesNotExist:
+            raise Exception("user not found")
```
XanderVertegaal: For consistency:

```suggestion
            raise Exception("User not found")
```

XanderVertegaal: You can also raise a `GraphQLError` instead of a general exception, which is a bit more precise.

In general, however, I would prefer the option of having a separate `errors` field on a Mutation and filling that with one or multiple error messages. (Check out [`DeleteCourseMutation` in DIAPP](https://github.com/CentreForDigitalHumanities/DIAPP/blob/develop/backend/course/mutations/DeleteCourseMutation.py) for an example.) Exceptions trigger GraphQL errors, which I would reserve only for unexpected cases where the mutation logic breaks due to some unforeseen circumstance. Here, you are already anticipating the possibility that a User might not exist for the provided ID, so in that case I would simply communicate that as part of a regular response. It is then up to the frontend to inspect the `errors` field and check whether it's empty.

backend/main/mutations/UserMutations.py
```diff
@@ -0,0 +1,55 @@
...
+        return UpdateUser(user=user)
+
+
+class DeleteUser(graphene.Mutation):
+    class Arguments:
+        id = graphene.ID(required=True)
+
+    success = graphene.Boolean()
+
+    def mutate(self, info, id):
```
XanderVertegaal: All mutations here are pretty powerful, and currently anyone with access to the API can perform them as often as they want. I know that this is just for testing/learning purposes and that we will delete this before it goes into production, but I would feel a bit safer if we only allowed these mutations to be performed by superusers.

```suggestion
@classmethod
def mutate(cls, root: None, info: ResolveInfo, id: str):
    current_user: User | AnonymousUser | None  = info.context.user
    if not current_user.is_superuser:
        return DeleteUser(success=false, errors=["User is not authorised to delete other users."])
```

frontend/pages/gql_tst.vue
```diff
@@ -17,17 +18,92 @@ const GET_USERS = graphql(`
...
+
+function submitUser() {
+  createUser({
+    email: email.value,
+    username: username.value,
+  }).then(() => {
+    refetch();
+    username.value = "";
+    email.value = "";
+  });
```
XanderVertegaal: You have to return the promise in order to not have it 'floating around'. Since you are raising exceptions in the backend, you also want to `catch` these.

The same is true for `refetch`, which you can mark with `void` to indicate that you don't care about the results, but see below for a better alternative.

```ts
function submitUser: Promise<void> {
    return createUser({
        ...
    }).then(() => {
        ...
    }).catch((error) => {
        // TODO: handle more gracefully
        console.error("Oops!", error);
    });
}
```

To go more modern, you can refactor this to use `async/await` syntax (which is easier to read, I feel), but this is up to you!

```ts
async function submitUser(): Promise<void> {
    try {
        const results = await createUser({
            email: email.value,
            username: username.value,
        });

        if (results?.data?.user) {
            console.log("User created!");
        }
    } catch (error) {
        console.error("Error creating user:", error);
        return;
    }

    username.value = "";
    email.value = "";
}
```

XanderVertegaal: You're probably calling refetch to get a new list of users. This works for this use case, but often you'll want to refetch other queries as well in components where you do not have access to those queries' `refetch` methods. For this reason, I don't really use `refetch` very much, but rather define a more general _cache eviction_ strategy. This tells Apollo which pieces of data should be removed from its cache. If the query is actively 'watched' by Apollo, then Apollo will notice that data is missing and automatically call refetch the query for you. You define a cache eviction strategy when you create the mutation (in `useMutation`), as follows:

```ts
const {
    mutate: createUser,
    loading: createUserLoading,
    error: createUserError,
} = useMutation<CreateUser>(CREATE_USER, {
    update: (cache) => {
        cache.evict({ fieldName: "users" });
        cache.gc();
    },
});
```

The field name corresponds to the name of the query field in the cache. Let me know if you'd like to learn how you can see what is in the Apollo cache. It's pretty neat!

XanderVertegaal (commented):
Nice work! 👌 It's clear that you're fast becoming a GraphQL master ;) 

Since you are keen to learn, I went full nitpick mode and have a lot of comments with tips/best practices etc. The only thing I really want to see addressed before this is merged is checking for user permissions (even if it's just a demo page with demo mutations).

tijmenbaarda:
It is not working correctly for 100%: when trying to create a user without the right permissions (i.e. not being superuser) no error is returned by the backend, and therefore no error is shown in the frontend:

afbeelding

I have tried to find a fix but without success...

tijmenbaarda:
What might be nice to add still is validation of the email address. Perhaps the `EmailValidator` from Django can be used for that.

EdoStorm96:
> It is not working correctly for 100%: when trying to create a user without the right permissions (i.e. not being superuser) no error is returned by the backend
> I have tried to find a fix but without success...

Ok, so I was a bit sloppy with how I implemented the errors for this. I did not do this correctly for the "not authorized" error. I have fixed this now.

I have also added validation of the email.

I do have a problem with logging in though ... I am not sure if this was already the case, but I cannot seem to get recognized as a superuser anymore ... @Xander, I will ask you about this tomorrow.

tijmenbaarda (approved):

EdoStorm96:
I've discussed a bit with @Xander, whether we should not just abandon this page ... Because I don't think it really needs to exist anymore.It was a nice exercise, but I'm going to open a PR which removes this page and related stuff instead.

PR #76, opened 2025-09-17
Feature/dev idp

### LINKED ISSUE #3: Add DevIDP to compose file

EdoStorm96:
fixes #3.

So here is the dev-idp containerized. I've also (with the generous help of @miggol) made it so that you can login and logout using the dev-idp. I don't know if we'll keep these settings ... I suppose they are okay for the dev profile and nothing more. It is a bit weird getting the containers to connect, but also sending users to the correct port ...

I've written a little instruction on how to connect this up in the README and updated the .env template and such.

The easiest way to check if login has worked is to login as admin and go to the admin panel.

I've now made it so that loggin in and out always returns to the homepage.

I hope you all will have some good suggestions :) 

PS. should the certs get committed? These ones I've just copied from the dockerized dev-idp repo, so it seems fine to me?


RoordinkFrank (approved):

.env.example
```diff
@@ -1,3 +1,5 @@
 DJANGO_SECRET_KEY="Change me!"
 DJANGO_ALLOWED_HOSTS=127.0.0.1,localhost,mr-django
-CSRF_TRUSTED_ORIGINS=http://127.0.0.1:5000,http://localhost:5000
\ No newline at end of file
+CSRF_TRUSTED_ORIGINS=http://127.0.0.1:5000,http://localhost:5000
+SAML_BASE_URL=https://localhost:8000/
+IDP_METADATA_URL=http://host.docker.internal:7000/saml/idp/metadata/
```
XanderVertegaal: This looks as if you're reaching the DevIDP from the Vue container through the host machine. I guess it doesn't matter much since we'll only be using this for dev, but since the Vue container and DevIDP are part of the same Docker network, they should be able to find each other without going to the host machine, provided the ports are properly exposed.

What happens if you use `http://dev-idp:7000` here?

README.md
```diff
@@ -190,3 +190,15 @@ is supposed to be created in development mode and to be committed to the codebas
...
 
 5. Open your browser and navigate to `http://localhost:5000` to visit the application!
+
+6. For the dev profile, you can login using the SAML [Development-IDP](https://github.com/CentreForDigitalHumanities/Development-IdP).
+
+   To make this work, you need to add MyResearch as a Service Provider to the Dev-Idp. To do so:
+      - visit `localhost:7000`
+      - Login as an admin user.
+      - Add a service provider with the following settings:
+         - Entity ID: `http://localhost:8000/saml/metadata/`
```
miggol: At this point I haven't gotten this to work with either yet, but we're for sure going to need this working on the `:5000` port in the future rather than `:8000`.

miggol (changes requested):
Couldn't get it configured :(

Same kind of issues as you experienced. IDP giving server errors, incorrect entity IDs etc. 

Need to put more effort into getting it running but for now I can at least request that we have communication with the backend happen on port 5000. On prod there's no other way to do it and SAML is one of those things where we want the environments to be as similar as possible.

miggol:
Solution for

```
Error during SAML2 authentication

ImproperlyConfigured
No active Service Provider object matching the entity_id 'http://localhost:5000/saml/metadata/' found

```

Make sure the Entity ID in the Dev-IDP matches the URL displayed in the error.

tijmenbaarda:
It mostly works for me, but I frequently get the following error after trying to log in:

```
Error during SAML2 authentication

UnknownSystemEntity
http://localhost:5000/saml/metadata/
```

This happens both before and after the SP is added. Logging in seems to have worked though, after refreshing the site.

tijmenbaarda:
When clicking on the metadata URL link, the user is referred to `http://host.docker.internal:5000/saml/metadata/`:

image

miggol (approved):

backend/myresearch/urls.py
```diff
@@ -32,3 +33,18 @@
         ),
     )
 ]
+
+# If SAML is enabled, add the required URL patterns for SAML
+if hasattr(settings, "SAML_CONFIG"):
```
XanderVertegaal: This may be a silly question, but why is this check necessary? Don't we always want to use SAML to login? Are we using another system?

backend/myresearch/settings/dev.py
```diff
@@ -30,3 +31,18 @@
         "in myresearch/settings/local_settings.py and Django will "
         "load them when running the dev profile."
     )
+
+# SAML STUFF
+
```
XanderVertegaal:
```suggestion
```

XanderVertegaal (approved):
It works for me, even if I specify `http://localhost:5000/saml/metadata/` as the remote metadata URL:

Image

backend/myresearch/urls.py
EdoStorm96: Yeah, I guess that makes sense!

EdoStorm96:
> It works for me, even if I specify `http://localhost:5000/saml/metadata/` as the remote metadata URL:
> Image

I notice your metadata is already loaded in? I think this overrides the Dev-Idp trying to load in the metadata from the remote url. I can change the remote url as well once I have created the Service provider (to any ol' random url even) but I cannot create the service provider with http://localhost:5000/saml/metadata as the remote metadata url. Did you do this or change it later, @XanderVertegaal?

PR #82, opened 2025-09-18
Add top_form property

XanderVertegaal:
@EdoStorm96 Since you were wondering how you can get from a deeply nested substep in a form to its top-level element, this is a computed property that does just that. It also features a test.


EdoStorm96 (approved):
Nice enhancement!

PR #83, opened 2025-09-22
Feature/links in frequent questions

### LINKED ISSUE #67: Feature/links_in_frequent_questions
No matter if we choose to have links in translations or in a separate composite link translations do not work properly if they come from a non static source.

RoordinkFrank:
Recent Activity has the following removed: 
import Recent Activity from "~/components/frontpage/RecentActivity.vue"; (importing itself)
Icons are removed.
Those two are still leftovers from feature/frontpage that had to be removed.
Other changes in Recent Activity are prettier only.

added feature:
Frequent questions can now handle links (and images, sort of). That way we can avoid putting html in translations. Ethics currently has multiple links in help_texts of questions. So setting this up now may also help later down the line.


frontend/components/frontpage/FrequentQuestions.vue
```diff
@@ -1,51 +1,91 @@
...
-  MockData.mockQuestions[props.questionsGroup];
+const questionAnswers: Record<string, AnswerPart[]> =
+    mockQuestions[props.questionsGroup];
+
+const isLinkData = (obj: LinkData | string | ImageData): obj is LinkData => {
+    return obj.hasOwnProperty("url");
+};
+const isImageData = (obj: LinkData | string | ImageData): obj is ImageData => {
+    return obj.hasOwnProperty("imageUrl");
+};
```
RoordinkFrank: Should I really be listening to prettier and change this
image
into this?
image

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
-  | "other";
+    | "processingRegistry"
+    | "ethicalCommission"
+    | "other";
 
-type MockQuestions = Record<MockQuestionKey, Record<string, string>>;
-
-export class MockData {
-  static mockQuestions: MockQuestions = {
+export type LinkData = {
```
RoordinkFrank: can the isLinkData and isImageData methods be part of the type in typescript? I do feel like they belong there but I am not sure how yet

XanderVertegaal: No, TypeScript types cannot have methods. You could use JS classes to combine methods and properties, but if you only have a single method per class, it is not really worth it I'd say.

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
+    text: string;
+    url: string;
+};
+export type ImageData = {
+    imageUrl: string;
+    altText: string;
+};
```
XanderVertegaal: Should these types live in `mockData.ts`? This file will be deleted at some point, but the types are necessary for your FrequentQuestions component. I suggest moving them there.

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
+export type Answer = (string | LinkData | ImageData)[];
```
XanderVertegaal: Union types are neat, but you need type guards (`isLinkData`, `isImageData`) in your component to narrow down the `Answer` type to the desired subtype (`LinkData`, `ImageData` or `string`). This is definitely possible, but since we are still prototyping, why not simply define a _single_ interface that your `FrequentQuestions` component can use? That makes checking it a lot easier with `v-if="answer.url"` vel sim.

```ts
// Not just Answer to avoid future clashes with actual form answers.
interface FrequentQuestionsAnswer {
    text: string;
    url: string | null;
    image: {
        src: string;
        altText: string;
    } | null;
}
```

frontend/locales/nl.json
```diff
@@ -29,7 +29,7 @@
   "Please Note: Accepted applications go to the archive. See above menu. The archive is available for everyone at the same facility with with a Solid-ID. " : "NB: Goedgekeurde aanvragen komen in het archief van deze portal te staan, zie het menu hierboven. Dit archief is toegankelijk voor iedereen met een Solis-ID die bij dezelfde faculteit werkt of studeert.",
   "UU-webpage of the FEtC-H": "UU-webpagina van de FETC-GW",
   "Regulations of the FEtC-H": "Reglement van de FETC-GW",
-  "Model documents": "Voorbeelddocumenten",
+  "Model documents": "Voorbeeld documenten",
```
```suggestion
  "Model documents": "Voorbeelddocumenten",
```

The original spelling without a space is correct.

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
     processingRegistry: {
-      "What is part of personal data?":
-        "Personal data is dat that tells something about a living person of whom you know the identity, from whom you can discern the identity or whom are otherwise recognizable. For example: contact information is personal data, but also a lot of research data is personal data. Measurement data, answers on a questionnaire, or interviews, observations, demographic data and much more. Data that can not be traced back to personal data are anonymous data, those are not personal data",
-      "Why are bananas curved?": "Because they are yellow.",
+        "What is part of personal data?": [
+            "Personal data is dat that tells something about a living person of whom you know the identity, from whom you can discern the identity or whom are otherwise recognizable. For example: contact information is personal data, but also a lot of research data is personal data. Measurement data, answers on a questionnaire, or interviews, observations, demographic data and much more. Data that can not be traced back to personal data are anonymous data, those are not personal data",
```
XanderVertegaal:
```suggestion
            "Personal data are data that tells something about a living person whose identity you know or can be discovered, or who is otherwise recognisable. Contact details are personal data, but data obtained through research (measurements, questionnaires, interviews, observations etc.) may count as personal data as well. Data that cannot be traced back to an individual are anonymous data. These are not considered personal data",
```

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
-      "Where can i find the documents I have to deliver?":
-        "During the application it may be possible that you need submit certain documents. Use the most recent model documents." +
-        " " +
-        "https://intranet.uu.nl/en/documents-ethics-assessment-committee-humanities",
-      "How long do responses take after submission?": "No idea.",
-      "What happens if my application gets rejected?":
-        "You will need to make a revision and resubmit the application.",
-      "Help my draft application suddenly changed":
-        "It is possible for the supervisor to change your application.",
+        "Where can i find the documents I have to deliver?": [
```
XanderVertegaal:
```suggestion
        "Where can I find the documents that I have to submit?": [
```

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
+            "During the application it may be possible that you need submit certain documents. Use the most recent model documents.",
```
XanderVertegaal:
```suggestion
            "During the application you may need to submit certain documents. Use the most recent model documents.",
```

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
+            {
+                text: "Model documents",
+                url: "https://intranet.uu.nl/en/documents-ethics-assessment-committee-humanities",
+            },
+            "You can also find the links to the documents during the application itself.",
+        ],
+        "How long do responses take after submission?": ["No idea."],
```
XanderVertegaal:
```suggestion
        "How long does it take to receive a reply after submitting?": ["No idea."],
```

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
+        "What happens if my application gets rejected?": [
+            "You will need to make a revision and resubmit the application.",
```
XanderVertegaal:
```suggestion
            "You will need to revise and resubmit the application.",
```

frontend/shared/mockData.ts
```diff
@@ -1,32 +1,51 @@
...
+        ],
+        "Help my draft application suddenly changed": [
```
XanderVertegaal:
```suggestion
        "Why did my draft application suddenly change?": [
```

frontend/shared/mockData.ts
XanderVertegaal, line 1: For readability reasons, please use whitelines in between all type declarations in this file.

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -1,132 +1,133 @@
...
 };
 const staticSendToProposal: Proposal = {
-  refNumber: 4234_2023,
-  title: "hi mom I am on the internet",
-  type: "proposal_in_ethics",
-  status: "SUBMITTED_TO_SUPERVISOR",
-  dateSubmitted: "2025-01-24",
-  lastEdited: "2025-04-23",
+    refNumber: 4234_2023,
+    title: "hi mom I am on the internet",
```
XanderVertegaal:
```suggestion
    title: "Placeholder",
```

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -1,132 +1,133 @@
...
+                <div class="text-muted mw-100">
+                    <div class="mw-100">
+                        {{ $t("Type") }}: {{ activity.type }}
+                    </div>
+                    <div class="mw-100">
+                        {{ $t("Status") }}:
+                        <template v-if="activity.status == 'DRAFT'"
+                            ><Pencil class="icon"
+                        /></template>
+                        <template
```
XanderVertegaal:
```suggestion
                        <template v-if="activity.status == 'DRAFT'">
                            <Pencil class="icon"/>
                        </template>
                        <template
```

frontend/components/frontpage/FrequentQuestions.vue
XanderVertegaal: It's technically safer to call prototype methods directly from the prototype itself and then apply it to the child object in question, so technically your suggestion to use `obj.prototype.hasOwnProperty.call(obj, "url")` works, but I agree it feels clunky.

A better way to check if a property exists in an object is to simply use `in`, so `if "url" in obj`. However, you cannot use `in` with a string, so you'll first have to exclude that you are working with a string:

```ts
    return typeof obj !== "string" && "url" in obj;
```

Have a look at my other comment in `mockData` about getting rid of a union type here. Maybe these checks are not necessary.

frontend/components/frontpage/FrequentQuestions.vue
```diff
@@ -1,51 +1,94 @@
...
 const props = defineProps<Props>();
 
-const questionAnswers: Record<string, string> =
-  MockData.mockQuestions[props.questionsGroup];
+const questionAnswers: Record<string, Answer> =
+    mockQuestions[props.questionsGroup];
+
+const isLinkData = (obj: LinkData | string | ImageData): obj is LinkData => {
+    return obj.hasOwnProperty("url");
+};
```
XanderVertegaal: Add a whiteline here.

frontend/components/frontpage/FrequentQuestions.vue
```diff
@@ -1,51 +1,94 @@
...
+    title: string;
 }
 const props = defineProps<Props>();
 
-const questionAnswers: Record<string, string> =
-  MockData.mockQuestions[props.questionsGroup];
+const questionAnswers: Record<string, Answer> =
+    mockQuestions[props.questionsGroup];
+
+const isLinkData = (obj: LinkData | string | ImageData): obj is LinkData => {
```
XanderVertegaal: Naming is hard but `obj` is really not a clear name. Please use `answer` or something slightly more informative :)

frontend/components/frontpage/RecentActivity.vue
XanderVertegaal: Funny placeholders are fine but at some point the project leaders are going to have a look at this website, and then it should look at least professional :)

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -1,132 +1,133 @@
...
 const recentActivity: RecentActivity[] = [
-  staticDraftProposal,
-  staticSendToProposal,
-  staticRoadmap,
+    staticDraftProposal,
+    staticSendToProposal,
+    staticRoadmap,
 ];
 type RecentActivity = Proposal | Roadmap;
 const isProposal = (obj: Proposal | Roadmap): obj is Proposal => {
```
XanderVertegaal: Add whitespaces in between each `const` and `type` here as well.

frontend/components/frontpage/FrequentQuestions.vue
```diff
@@ -1,51 +1,94 @@
...
+    type LinkData,
+    type MockQuestionKey,
+    type ImageData,
+    mockQuestions,
+} from "~/shared/mockData";
 
 interface Props {
-  questionsGroup: MockQuestionKey;
-  title: string;
+    questionsGroup: MockQuestionKey;
```
XanderVertegaal: `MockQuestionKey` is required for the `FrequentQuestions` component to work properly, so it should not be defined in `mockData.ts` or named `Mock...`. Please move its type definition here and rename it to `MockQuestionKey` or something else more appropriate.

frontend/components/frontpage/FrequentQuestions.vue
```diff
@@ -1,51 +1,94 @@
...
+                                    <img
+                                        :src="answerPart.imageUrl"
+                                        :alt="answerPart.altText"
+                                    />
+                                </template>
+                                <template v-else>
+                                    {{ $t(answerPart) }}
+                                </template>
+                                {{ " " }}
+                                <!-- I just want a space here there has to be a better way to do this. -->
```
XanderVertegaal: Why do you need a space here? I don't see any difference in the layout.

(Also, it is probably better if you simply ask one of us for help or add a comment to your PR in cases like this. If we had missed this comment during the review, it would have ended up on the server.)

frontend/locales/nl.json
```diff
@@ -61,5 +61,6 @@
   "Questions?": "Vragen?",
   "Additional Information": "Aanvullende informatie",
   "Previous": "Vorige",
-  "Next": "Volgende"
+  "Next": "Volgende",
+  "You can also find the links to the documents during the application itself." : "Je kan de links naar de documenten ook tijdens de applicatie vinden."
```
```suggestion
  "You can also find the links to the documents during the application itself." : "Je kan de links naar de documenten ook tijdens de aanvraag vinden."
```

XanderVertegaal (changes requested):
Glad to see the icons have been replaced with those from an external library :)

Here are some code-related comments. Let me know if anything is unclear!

frontend/locales/nl.json
RoordinkFrank: Are Model Documents still two words in that case?

frontend/components/frontpage/FrequentQuestions.vue
RoordinkFrank: This is done as well as it's indeed more logical to have it at FrequentQuestions. That said MockData and FrequentQuestions now have imports to each other. It is not circular but it doesn´t feel amazing either.

frontend/components/frontpage/FrequentQuestions.vue
RoordinkFrank: AnswerPart + " " + AnswerPart + " ". We made the decisions to not add a space to translations after a sentence but if you combine translations you need to read those.
Comment has been removed. I forgot to move the comment to a github comment after the pullrequest was made.

RoordinkFrank:
Only thing I have not been able to fix.
image
If ending a sentence with a link it's hard to end with a dot.

I think there may be a way to fix this but something tells me I will go into to much detail to be worth the effort.
A function could be made to analyse the pure text parts the answer. If the link text matches the word "Model Documents" then put the link in there.
Translations will look cleaner as well as the sentence is not interrupted in the translation so that would make them more reuseable.

I suppose such info would look like this
```
`{`
          text: "During the application you may need to submit certain documents. Use the most recent Model Documents",
          link: {
                       (link)text: "Model Documents"
                       url: null,
          },
          image: null,
},
```
Links that are not in a sentence are text: null


Still I may go to deep here so:

- [ ] Needs validation

edit: I just thought of a more simple solution. Move the { " " }} to only at the end a text instead of every answerPart. Add a single dot as text after the link and it should also be solved. Translations are still interrupted but this is much simpeler.

RoordinkFrank:
Another thing I have been thinking about is to add the option to add a SVG. If the link in question is a document we could add a small document SVG image just before the link to make it clear you are accessing a document directly.
Hopefully it ends up as something like
image


Still I may go to deep here so:

- [ ] Needs validation

frontend/components/frontpage/RecentActivity.vue
```diff
@@ -1,132 +1,140 @@
...
+    staticSendToProposal,
+    staticRoadmap,
 ];
+
 type RecentActivity = Proposal | Roadmap;
-const isProposal = (obj: Proposal | Roadmap): obj is Proposal => {
-  return obj.hasOwnProperty("status");
+
+const isProposal = (
+    recentActivity: Proposal | Roadmap,
```
XanderVertegaal:
```suggestion
    recentActivity: RecentActivity,
```

(also for the next function)

frontend/components/frontpage/FrequentQuestions.vue
```diff
@@ -1,51 +1,97 @@
...
-const questionAnswers: Record<string, string> =
-  MockData.mockQuestions[props.questionsGroup];
+export type FrequentQuestionKey =
+    | "processingRegistry"
+    | "ethicalCommission"
+    | "other";
+
+/** One or more FrequentAnswerParts form a FrequentAnswer.
+ * @text If url is not null then text is assumed to be the url text */
+interface FrequentAnswerPart {
```
XanderVertegaal: Thanks for implementing the interface, and good idea to add a docstring here! 👌

However, I don't think this approach with 'answer parts' combining into one answer is neater than what you had before. I see what problem you are trying to solve (having links within a text), but I would suggest we deal with such complex cases when we need to / when the user requests it. For a placeholder/demo component it is not really necessary.

My suggestion is to just use one `FrequentAnswer` object per answer, which is either some text, a link (url) or an image.

XanderVertegaal (changes requested):
Almost there! Check out my comment and let me know if there are any questions.

As for the two suggestions left in your comments (about parsing the `text` property of an answer and adding SVG files): let's not pursue these now. They are interesting ideas, but we can always consider adding those features if the users request it.

XanderVertegaal (approved):
Much better!

PR #85, opened 2025-09-25
Fix/remove gql tst

EdoStorm96:
Followup to #58.

Now that the site is getting more features, my proof-of-concept Graphql page can be removed :)

I think I got all of it? Maybe we'd like to keep the UserQuery, but with a different Queryset? But maybe that can be a new issue.


XanderVertegaal (approved):
I hope it was not too hard to kill your darlings. You will definitely be able to leverage what you learned here in future GraphQL types and mutations!

miggol (approved):
Time to take ol' Bessie round back with a shotgun

PR #86, opened 2025-10-02
Feat/setup permissions

### LINKED ISSUE #77: Create an overview of the required groups/permissions
Commit markdown file of permissions into repo, possibly somehwere in /docs/

### LINKED ISSUE #78: Think of a good system to implement permissions

EdoStorm96:
fixes #78

So here is my suggestion for how we handle permissions. I suggest:

- Every ObjectType (eg. `StudyType`) receives specific methods for receiving a queryset which a user can access if they have specific permission types (eg. edit or view). These are `viewable_queryset()` and `editable_queryset()`. These are called via the object type's `get_queryset()` method. This method also receives, alongside the usual stuff, a permission type from `MRPermissionTypes`.

- Every model receives (eg. `Study`) has specific methods for checking if a user has access to an instance of a model aka an object. These are `can_be_viewed_by()` and `can_be_edited_by()`. These are called via the model's `can_be_accessed_by()` method. This method receives a user and a permission type from `MRPermissionTypes`.

- For queries of multiple objects, the ObjectType's `get_queryset()` is used and for doing queries or mutations on single models, the `can_be_accessed_by()` method gets used. These two methods might need to get placed in a mixin, since I foresee duplicating these for every ObjectType and Model will get redundant, but I am not a 100% sure if that will always work, so for now I've just left these where they are.

To get a proof of concept to work I've implemented a really basic Study model. I've also made queries and mutations for studies. For now, this can be tested using the GraphQLView (localhost:5000/backend/api/graphql). It all works fine in my experience.

I have also added 2 groups, which we'll definitely need according to our stakeholders:
- Privacy officer
- fetc member
I've added a migration which automatically loads these groups. Be aware of that :)

I have not yet added the Study model or users which are members of these groups to create_dev_data, although that would be better ...

How I would test this is to create a couple of superusers, so you can login via the admin interface, make one of them a member of the privacy officer group via the admin interface and start making studies through mutations and then running some queries to see if they act as they should.

I think this design is pretty nice, but I think we need to carefully consider the naming as this will probably be re-used quite a bit.

I'm looking forward to hearing your feedback!

fixes #77
I also added a markdown table with the overview of permissions for the MVP and possible future permissions once we absorb the FETC.


EdoStorm96:
Oh and there is a problem with black? It can't parse create_dev_data, which I haven't toouched in this branch ...

backend/main/models.py
```diff
@@ -3,4 +3,25 @@
 
 
 class User(AbstractUser):
-    pass
+    @property
+    def is_privacy_officer(
+        self,
```
tijmenbaarda: Weird formatting! Is this Black's work?

backend/main/models.py
```diff
@@ -3,4 +3,25 @@
...
+    ):
+        return "Privacy officer" in [g.name for g in self.groups.all()]
```
tijmenbaarda: These strings could perhaps be declared as constants, like `PRIVACY_OFFICER = "Privacy officer"`, which can be reused in the migration as well.

backend/study/mutations/StudyMutations.py
```diff
@@ -0,0 +1,121 @@
...
+    @classmethod
+    def mutate(
+        cls,
+        root: None,
+        info: ResolveInfo,
+        title: str,
+        id: int,
+    ):
+
+        study = Study.objects.get(pk=id)
```
tijmenbaarda: This will cause an exception if not found; it won't return None. Use `try` ... `except Study.DoesNotExist`

backend/study/models.py
```diff
@@ -0,0 +1,43 @@
+from main.models import User, MRPermissionTypes
+
+from django.db import models
+
+
+class Study(models.Model):
+
+    created_by = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    title = models.CharField()
```
tijmenbaarda: Oh, I didn't know that it was possible to declare a CharField without a maximum length, but apparently it is :) It's only supported for postgresql and sqlite, but I suppose that's fine.

backend/study/queries.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        return queryset.all()
+
+    def resolve_my_study(root, info: ResolveInfo, id) -> Optional[Study]:
+        try:
+            study = Study.objects.get(id=id)
+        except Study.DoesNotExist:
+            return None
+        if study.can_be_accessed_by(info.context.user, MRPermissionTypes.EDIT):
+            return Study
```
tijmenbaarda: This is returning the type, I guess you meant `return study`

backend/study/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def can_be_accessed_by(self, user, permission):
+        """
+        Check whether an object can be accessed by a specific user, with a
+        specific permission type.
+
+        TODO: This function should probably live in some kind off mixin for all
+        models in MyResearch.
```
tijmenbaarda: Good for now, I think!

backend/study/models.py
```diff
@@ -0,0 +1,43 @@
...
+            f"{self}.can_be_accessed_by() has received a permission type which "
+            "has not yet been implemented."
+        )
+
+    def can_be_viewed_by(self, user):
+
+        if user == self.created_by:
+            return True
+        if user.is_privacy_officer:
+            return True
```
tijmenbaarda: Here only the privacy officer can see all studies, but in the list query (line 55 of `StudyType.py`) also the FETC members can see all studies, it seems. Also, according to the permissions doc file the privacy officer and the FETC members can do not have special permissions for seeing studies.

docs/MyResearchPermissions.md
```diff
@@ -0,0 +1,65 @@
+# Overview of Permssions in Myresearch
```
tijmenbaarda: Double dot in the filename

tijmenbaarda (changes requested):
I love the design! I still have to check whether it works properly by testing it with the some GraphQL queries, but here are some comments on the code.

backend/main/models.py
EdoStorm96: It is, but only because I put the comma's after self. I'll remove them.

backend/main/models.py
EdoStorm96: I don't really see how we can use the constant for the migration as it is now ... But if I changed the migration I could use them.

docs/MyResearchPermissions.md
```diff
@@ -0,0 +1,65 @@
...
+
```
RoordinkFrank, line 2: Adding permissions to the documentation is a neat idea and a separate table for current and possible future permissions works as well. It is however a bit hard to read right now. It seems to be possible to add limited css to .md files that hopefully will help.

```suggestion
<style>
    table th {
        background: grey;
    }
    table td {
        font-weight: bold;
        border: 1px solid grey;
    }
</style>

```

backend/study/mutations/StudyMutations.py
```diff
@@ -0,0 +1,118 @@
...
+            error = ErrorType(
+                field="id",
+                messages=["Study not found."],
+            )
+            return cls(errors=[error])
+
+        user = info.context.user
+
+        # Check if the user has Edit permission
+        if not study.can_be_accessed_by(user, MRPermissionTypes.EDIT):
```
RoordinkFrank: study.can_be_accessed_by(use, edittype) reads very clean. You immediately know what it does without having to read the method.

That said in this example could you not call can_be_edited_by directly? And if so what is the use case for can_be_accessed_by? Is it so the flow of the code matches get_queryset?

backend/study/types/StudyType.py
```diff
@@ -0,0 +1,66 @@
...
+            return cls.editable_queryset(user, queryset)
+        # NOTE: This should never happen.
+        raise ValueError(
+            f"{cls}.get_queryset() has received a permission type which has not "
+            "yet been implemented."
+        )
+
+    @classmethod
+    def viewable_queryset(
+        cls, user: User, queryset: QuerySet[Study]
```
RoordinkFrank: *edited*

This overlaps with can_be_viewed_by. I assume GraphQL is the reason that is has to be this way? Would it be possible to get these two if statements in a separate method?

backend/main/models.py
RoordinkFrank: I am not sure either what tijmen exactly means by migration but having Privacy officer as a constant would prevent spelling mistakes. The difference between Privacy Officer and Privacy officer is easily made after all.

backend/study/types/StudyType.py
```diff
@@ -0,0 +1,66 @@
...
+    ) -> QuerySet[Study]:
+        """
+        Filter the queryset based on view-permissions
+        """
+        if user.is_privacy_officer or user.is_fetc_member:
```
RoordinkFrank: The user.is_privacy_officer or user.is_fetc_member is fine in itself. It does depend how many roles we are expected to have. If this if statement grows to 10+ roles then it may be a problem.

An alternative would be to have check here if user.permissions[viewable_queryset] exists. And define the permissions in the user class. It would be a central place.

Edit: after thinking about it for a bit the current setup if far better if the goal is just the MVP. No need to over engineer things until they are needed.

RoordinkFrank (commented):
I do really like the naming of everything you have made so far, looks great. Commnents are in the code.

backend/study/mutations/StudyMutations.py
XanderVertegaal, line 1: I would split these up into individual files for individual mutations.

backend/study/queries.py
```diff
@@ -0,0 +1,64 @@
...
+
+from study.models import Study
+from study.types.StudyType import StudyType
+from main.models import MRPermissionTypes
+
+
+class StudyQuery(ObjectType):
+    study = Field(
+        StudyType,
+        id=Int(),
```
XanderVertegaal: For IDs, you can use GraphQL's `ID` type (import from `graphene`). Also, this argument should be marked as required. (My query fails without it.)


```suggestion
        id=ID(required=True),
```

backend/main/models.py
```diff
@@ -3,4 +3,26 @@
...
+    @property
+    def is_privacy_officer(self):
+        return self.PRIVACY_OFFICER in [g.name for g in self.groups.all()]
+
+    @property
+    def is_fetc_member(self):
+        return self.FETC_MEMBER in [g.name for g in self.groups.all()]
+
+
+class MRPermissionTypes(models.IntegerChoices):
```
XanderVertegaal: I would call it `MRPermission` (singular, as per [this suggestion](https://stackoverflow.com/questions/1335959/what-are-the-arguments-for-using-singular-vs-plural-for-enumerations)).

Is there any reason to use `IntegerChoices` instead of `TextChoices` here? The latter are more human-readable (esp. when browsing the DB).

backend/study/queries.py
```diff
@@ -0,0 +1,64 @@
...
+    my_study = Field(
+        StudyType,
+        id=Int(),
+    )
+
+    my_study_list = List(
+        StudyType,
+    )
+
+    def resolve_study(root, info: ResolveInfo, id) -> Optional[Study]:
```
XanderVertegaal: Decorate these resolvers with `@staticmethod`. That gets rid of the yellow squiggly (if you have one): "Instance methods should take a "self" parameter."

backend/study/queries.py
```diff
@@ -0,0 +1,64 @@
...
+    def resolve_study_list(
+        root,
+        info: ResolveInfo,
+    ) -> QuerySet[Study]:
+        queryset = StudyType.get_queryset(
+            Study.objects, info, permission=MRPermissionTypes.VIEW
+        )
+        return queryset.all()
+
+    def resolve_my_study(root, info: ResolveInfo, id) -> Optional[Study]:
```
XanderVertegaal: Instead of using 4 resolvers, it is shorter 2 and add an optional parameter `editable` to only see studies that you are allowed to edit.

XanderVertegaal: Also, you may want to type `id` as `id: str` (or `id: str | None` if it is an optional param).

backend/study/models.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Check whether an object can be accessed by a specific user, with a
+        specific permission type.
+
+        TODO: This function should probably live in some kind off mixin for all
+        models in MyResearch.
+        """
+        if not user.is_authenticated:
+            return False
+        if permission == MRPermissionTypes.EDIT:
```
XanderVertegaal: This would be a good place for a match/case statement.

backend/study/types/StudyType.py
```diff
@@ -0,0 +1,66 @@
...
+        queryset: QuerySet[Study],
+        info: ResolveInfo,
+        permission: int,
+    ) -> QuerySet[Study]:
+        """
+        Return the queryset of studies, based on the permission, defined in the
+        query.
+
+        TODO: This function should probably live in some kind off mixin for all
+        DjangoObjectType's in MyResearch.
```
XanderVertegaal:
```suggestion
        TODO: This function should probably live in some kind of mixin for all
        DjangoObjectTypes in MyResearch.
```

backend/study/models.py
```diff
@@ -0,0 +1,45 @@
...
+    title = models.CharField(
+        max_length=200,
+    )
+
+    def can_be_accessed_by(self, user, permission):
+        """
+        Check whether an object can be accessed by a specific user, with a
+        specific permission type.
+
+        TODO: This function should probably live in some kind off mixin for all
```
XanderVertegaal:
```suggestion
        TODO: This function should probably live in some kind of mixin for all
```

backend/study/types/StudyType.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @classmethod
+    def viewable_queryset(
+        cls, user: User, queryset: QuerySet[Study]
+    ) -> QuerySet[Study]:
+        """
+        Filter the queryset based on view-permissions
+        """
+        if user.is_privacy_officer or user.is_fetc_member:
+            return queryset
```
XanderVertegaal: I appreciate the `viewable_queryset` and `editable_queryset` methods, but now they contain the same logic as `User.can_be_viewed_by()` and `User.can_be_edited_by()`. When this model and the associated permission checks grow, it will be easy to change one and forget the other.

Is there a way you could define the logic only once and use it in both places?

backend/study/queries.py
```diff
@@ -0,0 +1,64 @@
...
+from main.models import MRPermissionTypes
+
+
+class StudyQuery(ObjectType):
+    study = Field(
+        StudyType,
+        id=Int(),
+    )
+
+    study_list = List(
```
XanderVertegaal: I don't mind `study_list`, but `studies` is slightly more conventional.

XanderVertegaal: Note that you are returning `List(StudyType)`, which means that both the `List` and its contents are nullable. Any consuming APIs/frontends will find it a lot easier to work with your query if you always return a list (empty or not) without nulls in it.

I recommend: `studies = List(NonNull(StudyType), required=True)`

docs/MyResearchPermissions.md
XanderVertegaal:
```suggestion
# Overview of Permissions in MyResearch
```

backend/study/queries.py
```diff
@@ -0,0 +1,64 @@
...
+        id=Int(),
+    )
+
+    my_study_list = List(
+        StudyType,
+    )
+
+    def resolve_study(root, info: ResolveInfo, id) -> Optional[Study]:
+        try:
+            study = Study.objects.get(id=id)
```
XanderVertegaal: Use `StudyType.get_queryset(Study.objects, info)` instead of `Study.objects`, so you do not miss any permission checks you might have in `StudyType.get_queryset()`.

(Feel free to disagree with me on this one. The DIAPP is really slow at the moment because of all the permission checks it needs to do, so there is definitely something to say about simplicity/brevity.)

backend/study/types/StudyType.py
```diff
@@ -0,0 +1,66 @@
...
+
+        TODO: This function should probably live in some kind off mixin for all
+        DjangoObjectType's in MyResearch.
+        """
+
+        user = info.context.user
+
+        if not user.is_authenticated:
+            return queryset.none()
+        if permission == MRPermissionTypes.VIEW:
```
XanderVertegaal: I like this 'signpost' version of `get_queryset()` that does not contain a lot of logic itself but refers to other methods for the actual checks!

(This would also be a good place for a match/case statement, btw.)

backend/study/mutations/StudyMutations.py
```diff
@@ -0,0 +1,118 @@
...
+        info: ResolveInfo,
+        title: str,
+    ):
+
+        user = info.context.user
+        # Initialize the study, but do not save()
+        study = Study(title=title, created_by=user)
+
+        # Check if the user has Edit permission
+        if not study.can_be_accessed_by(user, MRPermissionTypes.EDIT):
```
XanderVertegaal: I love this syntax :) It is so easy to read!

XanderVertegaal: This check is a bit silly atm, since you first create a study with `created_by=user` and then check whether `created_by` is in fact `user`. In the future you may want to check for a CREATE permission here or something like that :)

backend/study/mutations/StudyMutations.py
```diff
@@ -0,0 +1,118 @@
...
+            error = ErrorType(
+                field="", messages=["You are not authorized to create this study."]
+            )
+            return cls(errors=[error])
+
+        # If permission check gets passed, save() and return the study
```
XanderVertegaal: I don't think this comment is needed.

backend/study/mutations/StudyMutations.py
```diff
@@ -0,0 +1,118 @@
...
+    @classmethod
+    def mutate(
+        cls,
+        root: None,
+        info: ResolveInfo,
+        title: str,
+        id: int,
+    ):
+        try:
+            study = Study.objects.get(pk=id)
```
XanderVertegaal:
```suggestion
            study = StudyType.get_queryset(Study.objects, info).get(pk=id)
```

backend/study/mutations/StudyMutations.py
```diff
@@ -0,0 +1,118 @@
...
+    def mutate(
+        cls,
+        root: None,
+        info: ResolveInfo,
+        title: str,
+    ):
+
+        user = info.context.user
+        # Initialize the study, but do not save()
+        study = Study(title=title, created_by=user)
```
XanderVertegaal: The GraphQL endpoint is not authenticated, so `user` may be AnonymousUser. When I try to create a user without logging in, I see: `"Cannot assign \"<SimpleLazyObject: <django.contrib.auth.models.AnonymousUser object at 0x00000255E7457BF0>>\": \"Study.created_by\" must be a \"User\" instance."`

Perhaps it's good to do an early return if the user is `None` or `AnonymousUser`.

backend/study/mutations/StudyMutations.py
XanderVertegaal: This also goes for the DeleteMutation later on.

backend/main/models.py
```diff
@@ -3,4 +3,26 @@
 
 
 class User(AbstractUser):
-    pass
+
+    # Some constants used for group names
+    PRIVACY_OFFICER = "Privacy officer"
+    FETC_MEMBER = "FETC member"
```
XanderVertegaal, line 7: Could you put these constants in a TextChoices enum?

backend/main/migrations/0002_load_group_fixtures.py
XanderVertegaal, line 1: Interesting! I had no idea you could create dumps-as-migrations in this way. I'm not sure whether I am a fan of this. I think I'd rather keep fixtures/data separate from DB migrations. What do you think @miggol?

XanderVertegaal (changes requested):
Thanks for putting all the work in; it already works quite well and I love the human-readable syntax! <3

Most of my comments are nitpicks, but the one about reusing shared logic is one I would insist on. Feel free to fight me on anything else.

docs/MyResearchPermissions.md
EdoStorm96: I didn't know you can do that! However, I prefer the current styling ... So I'll leave this.

backend/study/mutations/StudyMutations.py
EdoStorm96: Calling the methods directly was my original idea ... But I found this cleaner, because you will only have to remember one method name, and then just pick the permission type you need.

backend/study/types/StudyType.py
EdoStorm96: Will work on this. Xander is also asking me.

docs/MyResearchPermissions.md
```diff
@@ -0,0 +1,65 @@
...
+|:------------------------------------|:-----------------------|:---------------------------|:------------------|:------------------|
+| 1. Studies                          |                        |                            |                   |                   |
+| 1.1 Lists                           | Access to own          | No access                  | Access to own     | Access to own     |
+| 1.2 Application detail              | Access to own          | No access                  | Access to own     | Access to own     |
+| 1.2 Create                          | Access to own          | No access                  | Access to own     | Access to own     |
+| 1.3 Update                          | Access to own          | No access                  | Access to own     | Access to own     |
+| 1.4 Delete                          | Access to own          | No access                  | Access to own     | Access to own     |
+| 1.5 View PDF                        | Access to own          | No access                  | Access to own     | Access to own     |
```
miggol: You've explained to me that these two tables represent the pre and post submission situations. That could be made clearer in this documentation.

backend/study/queries.py
EdoStorm96: But id is an int right?

EdoStorm96:
Alright, thanks for all your valuable feedback. I've made a major overhaul of the design, which does retain the nice syntax, but keeps permission logic in a centralized place. To achieve this, I've chosen to always reason from querysets.

I've made a custom manager, which has some new methods. The most important one being `accessible_objects()`. This method can be called like so for example: `Study.objects.accessible_objects(user, MRPermission.EDIT)` and will only return objects that the user has access to based on the permission type. I've made a base version of this manager in `main.utils.permission_utils`, which can be inherited from for specific objects, see `StudyManager` for an example of this.

I've also retained the `User.has_access_to()` method trough a little workaround, but it all uses the same logic. This method is not used currently, but can come in handy.

Talking to @XanderVertegaal and @miggol, we realized this PR does not really cover create permissions. I have made a placeholder for this (`Study.can_be_created_by`: it just checks if the user is authenticated.), but this should be implemented more extensively later.

From my testing it is all still working well, but the code has really fallen into place much nicer! I'm curious to hear your further thoughts!

backend/study/queries.py
XanderVertegaal: Django object IDs are `int`, but GraphQL IDs are treated as `str`, I believe. At least in DIAPP, where I tested this, `type(id)` yields `str`.

backend/myresearch/settings/generic_settings.py
```diff
@@ -26,6 +26,7 @@
 LOCAL_APPS = [
     "main",
     "form",
+    "study",
```
miggol: Now we're adding models I think it's worth having a talk about project structure, and how many distinct apps we want to end up with

backend/main/migrations/0002_load_group_fixtures.py
miggol: I'm a fan of using data migrations over fixtures wherever possible! Will elaborate in the meeting

backend/main/models.py
```diff
@@ -2,5 +2,34 @@
...
+    @property
+    def is_fetc_member(self):
+        return MRGroups.FETC_MEMBER in [g.name for g in self.groups.all()]
+
+    def has_access_to(self, object, mrpermission):
+        """
+        Utility function to check if a user has access to a specific object,
+        with a specific permission.
+        """
+        return object in object.__class__.objects.accessible_objects(self, mrpermission)
```
miggol: This could be (complemented by) `can_view(self, object,)` and `can_edit(self, object,)`

Personally I'd like to omit the opaque mrpermission argument wherever possible.

backend/main/models.py
```diff
@@ -2,5 +2,34 @@
...
+
+class MRPermission(models.TextChoices):
+    """
+    PermissionTypes are used in ObjectType's get_queryset method, to decide
+    which queryset gets returned or whether a user has access to an object,
+    based on the user's permissions.
+    """
+
+    VIEW = "View"
+    EDIT = "Edit"
```
miggol, line 37: I don't like mister permission. Passing around this extra argument, that can only have one of two values, to determine what you actually want to do? I would rather use two separate methods for viewing and editing wherever possible and forget about the argument. This makes for easier debugging as well.

For example, we could agree that `StudyType.get_queryset` just always means "view" by convention, and have a separate method for when we specifically want to access editable Studies.

Or we could annotate the `get_queryset()` results with an  attribute stating if each Study is editable by the requesting user, provided both methods assuming editable studies will always be a subset of visible studies.

I don't know if that makes Graphene sense though, perhaps `mrpermission` is the right way of doing things over in GraphQL land. But even then I would prefer integrating that into the info object, perhaps @XanderVertegaal knows if this is possible.

miggol (commented):
Just comments for now! Let's discuss this further in the meeting

backend/research/mutations/DeleteStudyMutation.py
```diff
@@ -0,0 +1,39 @@
...
+        root: None,
+        info: ResolveInfo,
+        id: int,
+    ):
+
+        try:
+            study = StudyType.get_queryset(
+                Study.objects,
+                info,
+            ).get(pk=id)
```
miggol: Shouldn't we be looking through the user's editable objects? Or doing some other kind of permissions check?

miggol: (same with UpdateStudyMutation)

miggol (changes requested):
I don't 100% remember everything we discussed last week, but this looks pretty much good to go! I did notice a lack of permissions check in the edit mutations.

Auto merge with develop works fine on my machine, but would be best to manually merge this branch first to be sure.

backend/research/mutations/DeleteStudyMutation.py
EdoStorm96: Really sorry about that. I've adressed it now. Thank you for finding it!

backend/research/mutations/CreateStudyMutation.py
```diff
@@ -0,0 +1,36 @@
...
+        root: None,
+        info: ResolveInfo,
+        title: str,
+    ):
+
+        user = info.context.user
+
+        # Check if the user has Create permission
+        if Study.can_be_created_by(user):
+            # Initialize the study, but do not save()
```
tijmenbaarda: I don't understand the comment `do not save()`

tijmenbaarda (approved):
Love the changes! One minor remark about a comment.

backend/research/mutations/CreateStudyMutation.py
EdoStorm96: That is outdated ... Will get rid!

PR #87, opened 2025-10-10
Feature/Form model and validation

XanderVertegaal:
This PR adds a composable (`useProcessForm`) to the frontend. This takes the `QueriedForm` object and creates two new objects: `FormWithValues` and `ValidationRules`.

1. `FormWithValues` is the exactly same as QueriedForm, but every question has two extra fields:
    - `value`: stores the value entered by the user. This field is bound to the value attribute of the question's corresponding `<input>`, `<textarea>` or `<select>` element. Now users can enter values and they will be stored in this object. The idea is that this will later be sent to the server to be processed.
    - `location`: a string that marks the location of the question within the form using the indices of the question, the substep and the step. This is needed for validation (see below).

2. `ValidationRules` is a validation object created with [Vuelidate](https://vuelidate-next.netlify.app/) (also used for DIAPP). The structure of this object mimics that of `QueriedForm` and provides validation errors for each question if the associated value is invalid. `ValidationRules` has an `$errors` array with `ErrorObject`s pointing to any invalid form values using a unique `$propertyPath` value. We use `question.location` to match errors from `ValidationRules` to questions in `FormWithValues`.

Now users can see when they have not provided a value for a required field, see below! Only the general `required` validator and the NumberQuestion-specific `positiveOnly` validator have been implemented so far, but others can be added relatively easily.

image

For frontend unit tests, I have added Vitest as a dev dependency. Run `npm run test` in `/frontend` to run tests.


frontend/composables/useProcessForm.ts
```diff
@@ -0,0 +1,255 @@
...
+                ...question,
+                location,
+                value: question.defaultValue,
+            };
+        case "FileUploadQuestionType":
+            return {
+                ...question,
+                location,
+                value: null,
+            };
```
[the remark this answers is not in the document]
XanderVertegaal: The type checker will complain if this switch statement is incomplete. A `default` block is not necessary here.

frontend/components/shared/FormWrapper.vue
```diff
@@ -1,44 +1,36 @@
...
 import { BSButton } from "cdh-vue-lib";
-import FormStepper, { type FormStepperConfig } from "./FormStepper.vue";
-import { type GetFormQuery } from "~/generated/gql/graphql";
-
-export type QueriedForm = NonNullable<GetFormQuery["form"]>;
-export type Step = NonNullable<QueriedForm["steps"][number]>;
-export type Substep = NonNullable<Step["substeps"][number]>;
-
-export type CombinedStep = Step | Substep;
+import type { QueriedForm } from "./FormWrapper";
```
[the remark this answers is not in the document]
XanderVertegaal: It imports it from a `.d.ts` file, so this is fine.

frontend/components/shared/FormStepper.d.ts
XanderVertegaal, line 1: The addition of `.d.ts` files is apparently necessary when you want to export types or interface and use in them in other components. It took me a _long_ time to figure out why I was getting so many type errors. Apparently, this is a recent change, so older projects like DIAPP did not have this problem (until we have to update it...).

frontend/components/shared/FormWrapper.vue
```diff
@@ -1,44 +1,36 @@
...
+const formObject = computed(() => formArtifacts.value.formWithValues);
+const validationRules = computed(() => formArtifacts.value.validationRules);
+
+const v$ = useVuelidate(validationRules, formObject, {
+    $autoDirty: true,
+});
+
+// Stepper configuration
+const formStepperConfig = computed<FormStepperConfig | null>(() =>
+    useBuildFormStepperConfig(props.queriedForm, props.currentStepSlug),
```
XanderVertegaal, line 29: There are no functional changes in `useBuildFormStepperConfig`. It's just a bit neater to keep that logic contained in its own composable. `FormWrapper` is getting quite large as it is.

frontend/composables/useBuildFormStepperConfig.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Build the configuration object required by the FormStepper component, based
+ * on the queried form and current step.
+ *
+ * @param queriedForm The queried form data.
+ * @param currentStepSlug The slug of the current step.
+ * @returns The FormStepper configuration.
+ */
+function useBuildFormStepperConfig(
```
EdoStorm96, line 12: Could this not be called buildFormStepperConfig? I find the 'use' a bit unintuitive.

frontend/components/shared/FileUploadQuestion.vue
```diff
@@ -1,33 +1,78 @@
...
 
-defineProps<Props>();
+interface Emits {
+    (e: "update:modelValue", value: File | null): void;
+}
+
+const props = defineProps<Props>();
+const emit = defineEmits<Emits>();
+
+const localQuestion = ref<File | null>(props.question.value);
```
EdoStorm96: So this refers to the currently uploaded file right? I would also prefer a different name for this (if I am reading this correctly). Maybe "currentFile"?

frontend/components/shared/FileUploadQuestion.vue
```diff
@@ -1,33 +1,78 @@
...
+            class="form-control"
+            :class="{ 'is-invalid': isInvalid }"
+            @change="onFileChanged($event)"
+        />
+        <div v-if="localQuestion" class="mt-2">
+            <strong>{{ $t("Selected file") }}:</strong>
+            {{ localQuestion.name }} ({{
+                (localQuestion.size / 1024).toFixed(2)
+            }}
+            KB)
```
EdoStorm96, line 68: Maybe lets do MB instead of KB? And the toFixed adds two decimals right? I don't think that is needed.

frontend/components/shared/FormWrapper.vue
```diff
@@ -1,44 +1,36 @@
...
-            active: substep.slug === props.currentStepSlug,
-            disabled: false,
-            // Let's only go 2 levels deep for now.
-            substeps: [],
-        })),
-    })),
-}));
-
-const allSteps = computed(() => getAllSteps(props.form));
+const formArtifacts = computed(() =>
```
EdoStorm96: I also find formArtifacts a bit obtuse ... How about ProcessedForm?

frontend/composables/useBuildFormStepperConfig.ts
EdoStorm96: I assume this is some vue/ts naming convention?

frontend/components/shared/MRForm.vue
```diff
@@ -26,20 +31,51 @@ const questionComponentMap = {
     TrueFalseQuestionType: SharedTrueFalseQuestion as Component,
     FileUploadQuestionType: SharedFileUploadQuestion as Component,
 };
+
+function getErrorsForControl(question: QuestionWithValue): ErrorObject[] {
```
EdoStorm96: What does `ForControl` mean? Would just `getErrors` also work?

frontend/composables/useProcessForm.test.ts
EdoStorm96, line 1: Very nice you added this!

frontend/composables/useProcessForm.ts
```diff
@@ -0,0 +1,255 @@
...
+        case "NumberQuestionType":
+            // Example: Add min/max value validation if needed
+            if (question.positiveOnly) {
+                rules.positiveOnly = helpers.withMessage(
+                    t("The number must be positive"),
+                    (value: number) => value >= 0,
+                );
+            }
+    }
+
```
EdoStorm96, line 252: It would be nice if you also implemented futureOnly here. I was confused why that was not showing up ...

EdoStorm96 (commented):
Another impressive one! Really good job, could not really find any mistakes as always ;p

As usual with TS stuff, i feel pretty out of my depth, so I may have missed things, but I looked over everything carefully and tried to understand it to the best of my knowledge. Many of my comments are about naming. Your variable names are not always intuitive to me ... This may be because of conventions I am not aware of, but maybe not.

I am having some doubts/questions about the design. So if I am reading all this code correctly, you are suggesting we keep track of the entire form, its values and its errors in the frontend. And it seems like you only want to send the data to the backend upon submission. Am I right in this assumption? At what point were you thinking we would send the data to the backend? That would explain why we need to keep track of the value and location for each question in vue.

This is an option I guess, which I am not entirely opposed to. But I am also not entirely convinced by it.

I would prefer to send data to the backend after filling out a single step. Then we only need to store the value of a question for a specific step, then send it back to the db, and then we reload the form with errors from the backend every time. Then we do not need the system you have implemented here for keeping track of question's locations. This would be my preference, because I feel like we need to keep track of the form in the backend anyway, and this PR just feels like we'd be keeping track of the form in two places.

We'd still keep a lot of this PR, but these features would really only be used for checking for errors on the current page, instead of for the whole form.

What do you think of this? I am as always open to be convinced otherwise and am curious what our other colleagues think of this. Let's discuss this at the dev meeting on wednesday.

frontend/composables/useBuildFormStepperConfig.ts
XanderVertegaal: I'm afraid so: all composables traditionally start with `use`. I agree it's not very intuitive/natural.

frontend/components/shared/FormWrapper.vue
XanderVertegaal: Better! 👍

frontend/components/shared/FileUploadQuestion.vue
XanderVertegaal: Done!

frontend/components/shared/FileUploadQuestion.vue
XanderVertegaal: Sure thing! I'll keep one decimal, though, since it will be good to see `0,4 MB` instead of just `0 MB`.

XanderVertegaal:
In response to @EdoStorm96's comment (for which thanks!):

This PR is against mostly meant as a proof of concept to show how we could track form values and errors in the frontend. It is intended to be agnostic when it comes to when/how often the frontend sends this data back to the backend. Still, you are definitely right that we should discuss this, so I'm glad you're bringing it up 😄 

To guide our discussion tomorrow, I think the following questions are most important.

- **When is form data submitted?** After every step? After every change to any value in the form? If we want the backend to update the `MRForm` by removing/adding steps based on what the user fills in, then frequent submissions are required.

- **Is a 'form' user-specific or general?** In other words: if the backend adds more steps, does this mean we need to have individual `MRForm`s? Do we save these objects to the DB or not?

- **Who is responsible for value-tracking and validation?** I think the backend should at least validate the data to make sure that it is correct and valid before saving it to the DB. The frontend should _know_ about errors so it can show to the user which fields are invalid, but whether it does any validation itself or simply receives error messages from the backend is up for debate.

XanderVertegaal:
Small update to this PR for future reviewers:

I have added `useFormState.ts`, which saves the form globally in such a way that you can safely navigate between steps and keep all values you have entered before. You can also inspect the globally stored `form-state` object in the 'Payload' tab of the dev tools. (I think you can even live edit the values there!)

image

PR #90, opened 2025-10-23
Pass current user information to frontend

tijmenbaarda:
This PR adds a Pinia store for the information about the currently logged in user, more or less like DIAPP does it, but with the important difference that in MyResearch we do not immediately refer to the login page if a user is not authenticated, because we allow users that are not logged in.


frontend/plugins/i18n.ts
```diff
@@ -20,8 +20,10 @@ export const i18n = createI18n({
     globalInjection: true,
     locale,
     messages,
-    formatFallbackMessages: true,
+    fallbackFormat: true,
```
tijmenbaarda, line 23: The name of this option was changed in the new ('composition') API of vue-i18n.

frontend/plugins/i18n.ts
```diff
@@ -20,8 +20,10 @@ export const i18n = createI18n({
...
     fallbackLocale: defaultLocale,
+    fallbackWarn: false,
+    missingWarn: false,
```
tijmenbaarda, line 26: Added these to suppress the warnings for missing strings in the 'en' locale, because these strings are missing by design. This also suppresses warnings for Dutch translations that are missing, but we have the CI checks for that.

frontend/components/layout/PageHeader.vue
```diff
@@ -48,11 +55,18 @@ function setLocale(locale: string) {
...
+                {{ $t("Welcome, {name}", { name: currentUserStore.currentUser?.fullName }) }} (<a
+                    href="/saml/logout"
+                    class="text-decoration-underline"
+                    >{{ $t("Logout") }}</a
+                >)
+            </div>
+            <div v-else class="ms-auto">
+                <a href="/saml/login" class="text-decoration-underline"
+                    >{{ $t("Login") }}</a
+                >
```
XanderVertegaal, line 78: I think we'll have to use NUXT_PUBLIC_SAML_URL here at some point.

frontend/components/layout/PageHeader.vue
```diff
@@ -9,6 +10,9 @@ function setLocale(locale: string) {
     i18n.locale.value = locale;
     localStorage.locale = locale;
 }
+
+const currentUserStore = useCurrentUserStore();
+await callOnce("user", () => currentUserStore.loadData());
```
XanderVertegaal, line 16: This only loads user data on application load. If the user logs in or out somewhere (e.g. in a different tab) the current app probably won't know about it until the user refreshes the page.

I used to think the DIAPP approach of checking for user data in middleware was overkill, but maybe it is also useful for MyResearch.

frontend/stores/current-user.ts
```diff
@@ -0,0 +1,39 @@
+import { defineStore } from "pinia";
+import { useApolloClient } from "@vue/apollo-composable";
+import { graphql } from "~/generated/gql";
+import type { GetCurrentUserQuery } from "~/generated/gql/graphql";
+
+interface State {
+    currentUser?: GetCurrentUserQuery["currentUser"];
+}
+
+export const useCurrentUserStore = defineStore("currentUser", {
```
XanderVertegaal: Since we're using the composition/setup API in our components, it would be nice to use the [setup syntax in Pinia stores](https://pinia.vuejs.org/core-concepts/#Setup-Stores) as well.

(This is just a stylistic thing, so feel free to disagree with me on this one, or leave it for a future issue!)

XanderVertegaal (commented):
Code-wise, this looks good. The login/logout URLs don't work yet, but we can leave that for another PR.

The only thing I'm not sure about it is whether checking for user data once at start-up is enough. It depends on how likely it is that useres will have multiple instances/tabs open of the same application where they can login or logout.

backend/main/types/UserType.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @classmethod
+    def get_queryset(
+        cls, queryset: QuerySet[User], info: ResolveInfo
+    ) -> QuerySet[User]:
+        return queryset.all()
+
+    @staticmethod
+    def resolve_full_name(parent: User, info: ResolveInfo) -> str:
+        return f"{parent.first_name} {parent.last_name}"
```
EdoStorm96: You can use [get_full_name()](https://docs.djangoproject.com/en/5.2/ref/contrib/auth/#django.contrib.auth.models.User.get_full_name) for this

frontend/components/layout/PageHeader.vue
```diff
@@ -27,7 +31,10 @@ function setLocale(locale: string) {
             <div class="ms-auto">
                 <!-- Spacer element, moves the next elements to the right -->
             </div>
-            <div v-if="true" class="border-left px-3">
+            <div
+                v-if="currentUserStore.currentUser?.isStaff"
+                class="border-left px-3"
+            >
                 <NuxtLink to="/" class="nav-link">
```
EdoStorm96, line 48: Could you make this link to the admin site while you're at it? :)

frontend/components/layout/PageHeader.vue
EdoStorm96: Yes, this might be good indeed.

EdoStorm96 (approved):
Looks good and seems to work well AFAICT! I have a few small remarks, but nice work here :)

frontend/components/layout/PageHeader.vue
tijmenbaarda: Changed this!

PR #94, opened 2025-11-12
Feature/create dev data studies

### LINKED ISSUE #95: Study model is not represented in create_dev_data
Because of the check in the `create_dev_data` command this command now fails immediately.

Fix is simple enough, just add Study to the list of models. But it might be nice to have the command check instead warn of a missing model, and have a Django test that fails when a model isn't represented in the mock data command.

EdoStorm96:
This was useful for implementing the lists. A small lil' PR ...

- I got the user fixture to correspond 1:1 to the users in the Dev IDP and added groups to them.
- I added a function for creating studies in create_dev_data. (I'm not sure if  we'll keep these study models ... But for now this is useful. Also the create_dev_data was complaining about this not being implemented. I tried to follow your lead on how to approach this @XanderVertegaal. I am not really sure if this should be an atomic transaction or not, but felt it probably should be?
- I found a bug when all objects had to get returned. Adding `.all()` to the return value of `_accessible_objects` fixes this.

Each bullet point corresponds to a commit. Happy reviewing :)

### DIFF
--- a/backend/main/management/commands/dev_fixtures/dev_users.json
+++ b/backend/main/management/commands/dev_fixtures/dev_users.json
@@ -11,7 +11,8 @@
             "is_staff": true,
             "is_active": true,
             "date_joined": "2023-05-02T08:27:53.470Z",
-            "is_superuser": true
+            "is_superuser": true,
+            "groups": [1,2]
         }
     },
     {
@@ -26,7 +27,8 @@
             "is_staff": false,
             "is_active": true,
             "date_joined": "2023-05-02T08:27:53.470Z",
-            "is_superuser": false
+            "is_superuser": false,
+            "groups": [1]
         }
     },
     {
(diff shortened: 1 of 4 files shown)


backend/main/management/commands/dev_fixtures/dev_users.json
```diff
@@ -35,13 +37,14 @@
         "fields": {
             "username": "professor2",
             "password": "plain$professor2",
-            "first_name": "Edo",
-            "last_name": "Storm",
-            "email": "E.Storm@harvard-example.edu",
+            "first_name": "Steve",
```
XanderVertegaal, line 40: I'm sorry Harvard fired you :(

backend/main/utils/permission_utils.py
```diff
@@ -34,8 +34,8 @@ def accessible_objects(self, user: User, mr_permission):
 
     def _viewable_objects(self, user: User):
         # Needs to be overwritten for a specific object's permissions
-        return self
+        return self.all()
```
XanderVertegaal, line 37: I found out the hard way myself (in DIAPP) that you need to call `.all()` or `.filter()` to turn a manager into a queryset.

XanderVertegaal (approved):
Good job! 👍 Everything seems to load in well, even with a flushed DB.

PR #97, opened 2025-11-20
Replace unittest with PyTest

### LINKED ISSUE #93: Pytest instellen met CI

tijmenbaarda:
Close #93.

This sets up pytest to replace unittest and I added it to the CI workflow. Tests can now be run by invoking the command `pytest` (inside or outside Docker).

PyTest is fully compatible with Python's standard unittest library, so tests don't have to be re-written, but for consistency and as an example I rewrote Xander's form unittests to pytest. @XanderVertegaal: I added you as a reviewer so that you can take a look at how I edited your tests, but feel free to ignore.

Note that I also upgraded the backend requirements (because I had to run `pip-compile`). This also upgrades some major releases, but since our current functionality is limited I think this is safe enough.


backend/requirements.txt
```diff
@@ -1,47 +1,43 @@
 #
-# This file is autogenerated by pip-compile with Python 3.12
+# This file is autogenerated by pip-compile with Python 3.11
```
miggol, line 2: Nice of you to compile requirements with the correct Python version!

miggol (approved):
Looks good!

PR #98, opened 2025-11-20
Feature/graphql list

### LINKED ISSUE #80: Create study list

EdoStorm96:
fixes #80.

So here is the GraphQLList PR :) We can now make nice lists. All the code is mostly taken from DIAPP. I highly recommend having a look at using_GraphQLList.md. It will help to make sense of what is going on in this PR for the unintiated.

For the initiated, I had to make a small change to the error handling in usePerformQuery. This method was suggested to me by Ty, but I don't fully understand it ...

Anyway, it is working well, so we can make a lot of lists! Let's go!


docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
+# Using GraphQLList
+
+For list pages in MyResearch, we use GraphQLList. This is a custom implementation of UUList, which is itself a custom vue component for rendering lists, part of [CDH Vue-lib](https://github.com/CentreForDigitalHumanities/Vue-lib). GraphQLList has been developed for use in the [DIAPP](https://github.com/CentreForDigitalHumanities/DIAPP). For now (nov. 2025), we have simply copied this implementation into MyResearch, but ideally, these vue components and custom Graphene/Django objects used for GraphQLList, should live in a separate library for other portals using the Vue/GraphQL/Django stack. This is a quick piece of documentation to explain how to use GraphQLList, as there are a lot of moving parts.
```
XanderVertegaal:
```suggestion
For list pages in MyResearch, we use GraphQLList. This is a custom implementation of UUList, which is itself a custom Vue component for rendering lists, part of [CDH Vue-lib](https://github.com/CentreForDigitalHumanities/Vue-lib). GraphQLList has originally been developed for use in the [DIAPP](https://github.com/CentreForDigitalHumanities/DIAPP). For now (Nov. 2025), we have simply copied this implementation into MyResearch, but ideally, these Vue components and custom Graphene/Django objects used for GraphQLList, should live in a separate library for other portals using the Vue/GraphQL/Django stack. This is a quick piece of documentation to explain how to use GraphQLList, as there are a lot of moving parts.
```

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+
+## What does GraphQLList do?
+
+GraphQLList produces a nicely styled list, based on a GraphQLQuery to the backend. It can be made searchable, have custom ordering and filters.
```
XanderVertegaal:
```suggestion
GraphQLList produces a nicely styled list, based on a GraphQL query to the backend. It can be made searchable and have custom ordering and filters.
```

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Preparing a Query
+
+Let's try to make a GQLList for a hypothetical `Book` object.
+
+To prepare a query, we firstly need to ensure that we are using our custom override of Graphene's DjangoObjectType: GQLListObjectType. This allows us to add fields to DjangoObjectType's Meta. In the ObjectType, we can also specify which fields should be searchable and what kind of filters we would like to have available for this object. We could for instance make the `Book`'s ObjectType look something like this:
```
XanderVertegaal:
```suggestion
To prepare a query, we first need to ensure that we are using our custom override of Graphene's DjangoObjectType: GQLListObjectType. This allows us to add fields to DjangoObjectType's Meta. In the ObjectType, we can also specify which fields should be searchable and what kind of filters we would like to have available for this object. We could for instance make the `Book`'s ObjectType look something like this:
```

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+        cls,
+        queryset: QuerySet[Book],
+        info: ResolveInfo,
+    ) -> QuerySet[Book]:
+        return queryset
+
+```
+This ensures that the `Book`'s Title and Author are searchable and the we can filter the list of books based on its color (supposing this is a related object in this case). For more information on how to write filters, check out the docs for [django-filter](https://django-filter.readthedocs.io/en/stable/index.html).
+
+Next, we need to add a `book_pages` query to our `BookQuery` object. This uses a custom field object, which adds some relevant attributex to a list-like query for pagination. The field is called GQLListPaginationConnectionField. Add the field to your query object like so:
```
XanderVertegaal:
```suggestion
Next, we need to add a `book_pages` query to our `BookQuery` object. This uses a custom field object, which adds some relevant attributes to a list-like query for pagination. The field is called `GQLListPaginationConnectionField`. Add the field to your query object like so:
```

backend/research/queries.py
```diff
@@ -16,13 +17,16 @@ class StudyQuery(ObjectType):
     )
 
     studies = List(
-        NonNull(
-            StudyType,
-        ),
+        StudyType,
```
XanderVertegaal: Why remove the `NonNull` here? Now this query's return is typed as `[StudyType]!`, which means that it returns a list of `StudyType` and `None` values.

Unless you have a good reason to introduce `None` values in the returned list, I would advise against this. You will hardly ever need to pass `None` values in a list, and it avoids having to do a null check in the frontend.

backend/research/queries.py
```diff
@@ -3,6 +3,7 @@
 
 from django.db.models import QuerySet
 
+from api.graphql.pagination_field import GQLListPaginationConnectionField
 from research.models import Study
 from research.types.StudyType import StudyType
 from main.models import MRPermission
```
XanderVertegaal: Unused import.

backend/research/queries.py
```diff
@@ -16,13 +17,16 @@ class StudyQuery(ObjectType):
...
-        NonNull(
-            StudyType,
-        ),
+        StudyType,
         mr_permission=String(required=True),
         required=True,
     )
 
+    study_pages = GQLListPaginationConnectionField(
+        StudyType,
```
XanderVertegaal, line 27: The lack of a `NonNull` here causes this query to have the same return type as the one noted in the comment above. Again: it's not terrible, but slightly more annoying to deal with in the frontend.

Here, however, adding `NonNull` breaks `GQLListPaginationConnectionField`, so this is something to keep in mind. It would be great if we could improve `GQLListPaginationConnectionField` (also for DIAPP!) so that it supports `NonNull` (or just treats whatever type you enter into it as such by default). This is not something for this PR, however! It's fine for now to keep this as it is :)

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+```
+    book_pages = GQLListPaginationConnectionField(
+        BookType,
+    )
+```
+With a resolver, that will look similar to other List resolvers:
+```
+    @staticmethod
+    def resolve_book_pages(root, info: ResolveInfo, **kwargs) -> QuerySet[Book]:
+        return BookType.get_queryset(book.objects, info)
```
XanderVertegaal:
```suggestion
        return BookType.get_queryset(Book.objects, info)
```

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+```
+This is all the work in the backend done!
+## Rendering your query using GraphQLList
+Next we'll render this Query using our GraphQLList component. First, create a new page where you want your list to live. Then we'll need to setup a few things. The most important thing is your graphql query, which might look something like this:
```
XanderVertegaal:
```suggestion
Next we'll render the result of this query using our GraphQLList component. First, create a new page where you want your list to live. Then we'll need to setup a few things. The most important thing is your GraphQL query, which might look something like this:
```

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+```
+const GET_STUDY_PAGES = graphql(`
```
XanderVertegaal: It would be nice if you could stick to the `Book` example in the previous section for continuity :)

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+    }
+    pageInfo {
+      count
+      limit
+      offset
+    }
+  }
+}
+```
+There is some extra variables here, which are provided via GQLListPaginationConnectionField. Furthermore, there is also the variable that we use for our filter (`colorIds`). Our `Book` objects are withing the result part of the query.
```
XanderVertegaal:
```suggestion
There is some extra variables here, which are provided via GQLListPaginationConnectionField. Furthermore, there is also the variable that we use for our filter (`colorIds`). Our `Book` objects are available in the `result` property of the query result.
```

docs/using_GraphQLList.md
```diff
@@ -0,0 +1,227 @@
...
+
+Next, we'll need to define the variable that we are passing into our GraphQLList component. These are kindoff like default variables for search and ordering. In our case they might look like this:
```
XanderVertegaal:
```suggestion
Next, we'll need to define the variables that we are passing into our GraphQLList component, such as `search` and `ordering`. You can provide default values for them in the following way.
```

backend/main/management/commands/create_dev_data.py
```diff
@@ -76,6 +76,7 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument("--force", action="store_true")
         parser.add_argument("--silent", action="store_true")
+        parser.add_argument("--ignore-missing-models", action="store_true")
```
XanderVertegaal, line 79: Nice :) I'm glad there is a bypass for this check. Not all models need to have dev data if you just want some dummy data to play around with.

I'm actually not sure `create_dev_data` should be responsible for checking whether all models are included. Maybe the `check_all_models_implemented()` is actually best outfactored to an independent test to have a good separation of concerns. We can then add it to `pr-checks.yml` and have it run on every opened/updated PR.

backend/main/queries.py
```diff
@@ -20,3 +25,10 @@ def resolve_current_user(root: None, info: ResolveInfo) -> User | None:
         except User.DoesNotExist:
             # Should never happen.
             return None
+
+    @staticmethod
+    def resolve_users(root: None, info: ResolveInfo) -> list[User] | list:
```
XanderVertegaal: The last line returns a queryset, not a list.

```suggestion
    def resolve_users(root: None, info: ResolveInfo) -> QuerySet[User]:
```

For line 33 I'd use `return User.objects.none()` which is consistent and semantically clearer.

backend/main/queries.py
```diff
@@ -20,3 +25,10 @@ def resolve_current_user(root: None, info: ResolveInfo) -> User | None:
...
+        user = info.context.user  # type: User
```
XanderVertegaal: Just FYI, in DIAPP I'm moving away from ` user = info.context.user  # type: User` in favour of `user: User = info.context.user`

It's functionally equivalent but looks just a bit cleaner in my opinion. Adopt it if you like it too, but feel free to ignore it :)

docs/using_GraphQLList.md
```diff
@@ -57,22 +57,21 @@ With a resolver, that will look similar to other List resolvers:
...
+query GetBookPages(
   $limit: Int
   $offset: Int
   $ordering: String
   $search: String
   $colorIds: [ID]
 ) {
-  studyPages(
-    mrPermission: "View"
+  BookPages(
```
XanderVertegaal: `bookPages` (camelCase)

frontend/components/study_detail/StudyDetailsSidebar.vue
```diff
@@ -0,0 +1,44 @@
...
+            aria-expanded="false"
+        >
+            {{ $t("Show sidebar") }}
+        </button>
+        <div id="exampleSidebar" class="uu-sidebar-collapse collapse">
+            <h3>{{ $t("Creator details") }}</h3>
+            <ul>
+                <li class="mt-2">
+                    {{ $t("Created by:") }}
+                    {{ study?.createdBy.fullName }}
```
XanderVertegaal, line 26: Isn't this going to show `undefined` if no study is found?

You can coerce this to an empty string or some other fallback message: `{{ study?.createdBy.fullName ?? "" }}`

Alternatively, you can choose not to show this component at all if there is no study available: remove the `| null` from the type of the `study` prop and add a `v-if` condition in the parent component. That keeps this component a bit cleaner.

```ts
defineProps<{
    study: StudyType;
    study: GetStudyQuery["study"];
    randomDatePastYear: string;
}>();
```

frontend/locales/nl.json
```diff
@@ -62,9 +62,39 @@
...
+  "This field is required": "Dit veld is verplicht",
+  "The number must be positive": "Het getal moet positief zijn",
   "Welcome, {name}": "Welkom, {name}",
   "Logout": "Uitloggen",
   "Login": "Inloggen",
+  "You can also find the links to the documents during the application itself.": "Je kan de links naar de documenten ook tijdens de aanvraag vinden.",
+  "Study": "Studie",
+  "Study overview": "Studieoverzicht",
+  "Creator details": "Details aanvrager",
+  "Created by:": "Aangemaakt door:",
```
XanderVertegaal: I'd suggest to keep punctuation marks (i.c. ':') out of the translated strings. They are language-independent and including them here makes these strings difficult to reuse.

XanderVertegaal (commented):
Here are some more nitpicks :)

frontend/pages/studies/index.vue
```diff
@@ -0,0 +1,151 @@
...
+                                {{ row.title }}
+                            </NuxtLink>
+                        </td>
+                        <td>
+                            {{ row.createdBy.fullName }}
+                        </td>
+                    </tr>
+                    <tr v-if="isLoading">
+                        <td colspan="5" class="text-center">
+                            <Loading />
```
tijmenbaarda, line 140: This component does not exist and this causes warnings in the frontend logs; I think you may just copy it from DIAPP as well!

tijmenbaarda (approved):
Very nice! It works well, apart from the frontend error because of a missing component (see my comment in code).

What is still missing is error handling with a visible message to the user, but I think that may go in a separate PR (I could pick that up next).

I would suggest that you add a comment to all files that were (mostly) copied from DIAPP.

Issue #71, opened 2025-09-11
Commit schema.json to repo


miggol:
This allows production containers to run without having to codegen.

Requires a workflow for checking the `schema.json` is up to date in every PR

PR #99, opened 2025-11-24
Feature/prod fixes

miggol:
Please admire the product of multiple hours of swearing: no more than 14 lines of config changes. Committing the schema does pad my stats for this repo nicely, though.

Note that I think the incorporation of the GraphQL schema into the repo can be a lot nicer and I don't consider this a fix for #71. It's just a bandaid so we can have a working demo on https://dev.myresearch.hum.uu.nl/ .

To that same end, the dev server is now tracking this branch.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -6,7 +6,7 @@ services:
       - POSTGRES_PASSWORD=postgres
     healthcheck:
       test: 'psql -c "\l" postgres postgres'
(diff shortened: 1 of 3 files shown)


tijmenbaarda:
Looks good! It is not completely clear to me if you intend to merge this into develop or not, since you mention that #71 would make this fix obsolete.

PR #100, opened 2025-11-25
Feature/permission tests

EdoStorm96:
Here are some simple tests for the current permissions.

they pass so that is good.

I have one small issue. I cannot access the groups that are loaded in migration `0002_load_group_fixtures`. Therefore I recreate them. I tried a lot of things but I cannot seem to get to them in any way. To me, recreating them is not a big deal, as I remake them using the same constants that are used throughout the app, so as long as these are maintained, the tests will work.

If anyone has a suggestion how I could access the `Group` objects that are created in the migration, let me know. I tried a lot of things, but could not get it to work sadly ...


backend/research/tests.py
```diff
@@ -1,3 +1,126 @@
...
+        self, test_fetc_user: User, test_study: Study, test_po_study: Study
+    ):
+        """Test that fetc_user can view all studies and edit no studies (as
+        they have not created any)"""
+
+        viewable_studies, editable_studies = self._get_editable_and_viewable_studies(
+            test_fetc_user
+        )
+
+        assert set(viewable_studies) == {test_study, test_po_study}
```
tijmenbaarda, line 120: Cool, I didn't know this syntax for creating a set

backend/research/tests.py
```diff
@@ -1,3 +1,126 @@
...
+def test_po_study(test_po_user) -> Study:
+    return Study.objects.create(title="test_po_user's study", created_by=test_po_user)
+
+
+####################
+# Permission Tests #
+####################
+
+
+@pytest.mark.django_db(transaction=True)
```
tijmenbaarda: `transaction=True` is not needed here, it seems

backend/research/tests.py
```diff
@@ -1,3 +1,126 @@
...
+
+
+@pytest.fixture
+def test_anonymous_user() -> AnonymousUser:
+    return AnonymousUser()
+
+
+@pytest.fixture
+def test_po_user() -> User:
+    po_user = User.objects.create(username="po_user", password="1234")
```
tijmenbaarda: I don't know if there is any difference, but there is the method `User.objects.create_user` for this.

backend/research/tests.py
```diff
@@ -1,3 +1,126 @@
...
+from main.models import MRGroups, MRPermission, User
+from .models import Study
+
+#################
+# User fixtures #
+#################
+
+
+@pytest.fixture
+def test_anonymous_user() -> AnonymousUser:
```
tijmenbaarda: Maybe avoid the `test_` prefix for fixtures to avoid the suggestion that these are tests

tijmenbaarda (approved):

PR #101, opened 2025-11-26
Fix/accordion collapse

### LINKED ISSUE #53: fix/accordion_collapse
Accordion does not collapse properly in FrequentQuestions component.
The basic bootstrap example on the official website has the exact same problem inside myresearch but functions fine in example code sites.
In debug mode you can see that the accordion actually collapses but immediately gets overridden by something else.

RoordinkFrank:
fixes #53

The example bootstrap accordion code did not work on myresearch but did work on DIAPP. This copies the plugins/bootstrap.ts from diapp.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -64,11 +64,10 @@
             }
         },
         "node_modules/@apidevtools/json-schema-ref-parser": {
-            "version": "14.1.1",
-            "resolved": "https://registry.npmjs.org/@apidevtools/json-schema-ref-parser/-/json-schema-ref-parser-14.1.1.tgz",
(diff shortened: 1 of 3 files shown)


tijmenbaarda (approved):
Works perfectly!

We could think of only activating the modules that we are actually using, but personally I don't object to including them all.

RoordinkFrank:
> Works perfectly!
> 
> We could think of only activating the modules that we are actually using, but personally I don't object to including them all.

Not importing what you don´t use is definitely good.  There is a risk though that we simply have forgotten this file exists by the time we add a new bootstrap component and we are stuck wondering why it won´t work. I am not sure what is more important.

XanderVertegaal:
I would be in favour of just importing what we need, not everything the DIAPP happens to import.

PR #104, opened 2025-12-01
Run npm install on frontend container start

tijmenbaarda:
This fix solves the problem that npm dependencies were not automatically installed when `package.json` was altered. As Michael suggested, the dependencies are now installed as the first step of the `start_container.sh` script, so this happens every time the container starts. The dependencies are no longer installed during image building. This did not work, because the `node_modules` directory was mounted as a volume and as such they were not part of the image.

Note that the backend still installs the dependencies while the image is being built: here the dependencies are installed as part of the Python installation, which lives in the image. I have documented this difference in the README file.

I updated the Bootstrap dependency to the latest point release to test this fix.

### DIFF
--- a/frontend/Dockerfile.dev
+++ b/frontend/Dockerfile.dev
@@ -9,14 +9,7 @@ RUN apk add python3
 # Set the working directory in the container.
 WORKDIR /app
 
-# Copy package.json and package-lock.json (if available).
-COPY package*.json ./
-
-# Copy start_dev.sh script
-COPY start_dev.sh ./
-
-# Install project dependencies
-RUN npm install
+# Source is not copied because it is mounted as a volume
 
 # Nuxt 3 default port is 3000.
 # Does not actually expose the port but documents it.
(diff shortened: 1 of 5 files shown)


frontend/Dockerfile.dev
```diff
@@ -9,14 +9,7 @@ RUN apk add python3
 # Set the working directory in the container.
 WORKDIR /app
 
-# Copy package.json and package-lock.json (if available).
-COPY package*.json ./
-
-# Copy start_dev.sh script
-COPY start_dev.sh ./
```
tijmenbaarda, line 16: I also removed copying the `start_dev.sh` script, because this is part of the source directory which is being mounted. This works, but maybe I am overlooking something.

RoordinkFrank (approved):
branch develop:
 sudo npm remove lucide-vue-next
sudo docker compose --profile dev up --build -d
changes are not detected 
 sudo docker compose --profile dev down
gh pr checkout 104
sudo docker compose --profile dev up --build -d

Image

 sudo npm install lucide-vue-next
sudo docker compose --profile dev down
sudo docker compose --profile dev up --build -d

Image

No errors. It all seems to work.

Issue #102, opened 2025-11-27
Add BaseTriggerValue model


XanderVertegaal:
Step/question conditions have a field 'trigger value', that determines when the condition should 'activate' and hide or show questions or steps.

In the current implementation, this trigger value is a generic JSONField, but for more flexibility, we'd have a M2MField, ForeignKey or maybe just NumberField, CharField and BooleanFields, depending on the trigger question, e.g. if the question is a BooleanQuestion, a trigger associated with it would have its trigger value stored in a BooleanField.

Based on @miggol's suggestions, we can implement a BaseTriggerValue (or something like that) model that is inherited by specific subtypes that provide a specific implementation of the 'value' field. This structure would be similar to that of 'BaseQuestion' and its associated TrueFalseQuestion, TextQuestion etc.

PR #105, opened 2025-12-02
Feature/form conditions

### LINKED ISSUE #64: Model form values in frontend
Using a random generated by `create_dev_data`, the frontend should be able to record data and send it back to the backend.

No DB saving or validation required yet.

### LINKED ISSUE #89: Decide on whether Steps should actually exist
We seem to have reached agreement on questions being singletons, while question-answer pairs are unique to each study revision, bound together by some kind of Document object.

We have yet to decide how Steps will be handled: Do they exist in the database or are they ephemeral? Are they singletons or per-study? Much to discuss

XanderVertegaal:
This PR allows us to define conditions on steps and questions in the form structure. It's another big one with lots of moving parts, so here is some additional documentation for your reviewing pleasure.

### New models
 - `QuestionCondition`/`StepCondition`: these models are used to register that one question ('trigger') influences how another question ('target') is represented in the form. There are currently four possible effects (`condition_type`) which specify what should happen to the target question if the condition is met. The condition is considered to be met if the value in `trigger_value` (a JSONField) matches that of the user's answer to the trigger question (i.e. QuestionResponse.answer, also a JSONField).
   - These models both inherit from `BaseCondition`, now implemented as an abstract base class. I'm still not sure this is a better pattern to follow than the multi-class inheritance pattern used for the various `Question` models, so if you have any ideas about this, I'd be happy to hear them!
 - `QuestionResponse`: records a user's answer to a question, stored in a JSONField.
 - `UserFormSubmission`: binds QuestionResponses together along with some metadata.

### User-specific forms
With this PR, the `MRForm`, `Step` and `Question` objects in the database are now considered templates and not queryable anymore. Instead, the frontend now receives a user-specific `UserFormType` object, which contains the user's previously given answers, omits hidden questions etc.

Whenever a user queries a form, the backend first collects the corresponding template (`MRForm`). It will then first create a `FormEvaluator`, which determine whether questions are visible, whether steps should be repeated etc. based on previous user input and the provided form config/template. This evaluator is then passed to a `UserFormResolver` which converts everything into GraphQL types so it can be exported. The split between an 'evaluator' and a 'resolver' is not necessary, but it separates the logic layer from the presentation layer (and breaks up the code a bit).

Because the backend does all this heavy lifting/modification of the form, the required changes to the frontend are relatively minor.

### Cool, but how do I test this?

The frontend does not send anything back to the backend yet, so the frontend form is still static, unfortunately. (There is no mutation for submitting a form in place yet.) For now, I recommend visiting the Django Admin, adding `QuestionCondition`s and `QuestionResponse`s there and seeing how that changes how the form is represented in the frontend.

### What remains to be done?
- Get better typing for `trigger_value` using an inherited base model (#102).
- Write a mutation that submits the data and creates new `QuestionResponse` and `UserFormSubmission` objects.
- Make sure the frontend sends back the form frequently so the backend can recalculate what should be shown/hidden: at least after every step, probably also after every question that has a trigger attached to it.

Closes #89 (obsolete), closes #64


backend/api/graphql/schema.py
```diff
@@ -25,14 +24,6 @@ class Query(FormQueries, UserQueries, StudyQuery, ObjectType):
...
-    create_study = CreateStudyMutation.Field()
-    update_study = UpdateStudyMutation.Field()
-    delete_study = DeleteStudyMutation.Field()
-
-
-class Mutation(ObjectType):
-
     create_study = CreateStudyMutation.Field()
     update_study = UpdateStudyMutation.Field()
     delete_study = DeleteStudyMutation.Field()
```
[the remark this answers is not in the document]
XanderVertegaal: This used to be the case, but not anymore.

backend/form/admin.py
```diff
@@ -1,3 +1,428 @@
...
+                "description": "For 'repeat' conditions: set repeat_count. For 'repeat_dynamic': check use_answer_as_count.",
+            },
+        ),
+    )
+
+    def trigger_value_preview(self, obj):
+        value_str = str(obj.trigger_value)
+        return value_str[:30] + "..." if len(value_str) > 30 else value_str
+
+    trigger_value_preview.short_description = "Trigger Value"
```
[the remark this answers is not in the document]
XanderVertegaal: The translated fields are registered in translation.py.

backend/form/types/QuestionType.py
```diff
@@ -1,159 +1,249 @@
...
+        required=True, description="The ID of the question this is based on."
+    )
+    repeat_index = Int(required=True)
+    answer = JSONString()
+
     text_nl = String(required=True)
     text_en = String(required=True)
     description_nl = String(required=True)
     description_en = String(required=True)
     required = Boolean(required=True)
```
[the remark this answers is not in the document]
XanderVertegaal: These fields are correctly defined in translation.py

backend/form/types/QuestionType.py
```diff
@@ -1,159 +1,249 @@
...
+
+    @staticmethod
+    def resolve_multiple(parent, info: ResolveInfo):
+        if hasattr(parent.question, "selectquestion"):
+            return parent.question.selectquestion.multiple
+        return parent.question.multiple
+
+    @staticmethod
+    def resolve_options(parent, info: ResolveInfo):
+        if hasattr(parent.question, "selectquestion"):
```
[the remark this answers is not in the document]
XanderVertegaal: This is true, but optimizing this now would be too premature.

backend/api/graphql/schema.py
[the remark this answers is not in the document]
EdoStorm96: Woops, My bad I think ...

backend/form/services/form_evaluator.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def __init__(self, form: MRForm, user: UserType, create_submission: bool = False):
+        self.form = form
+        self.user = user
+        self.create_submission = create_submission
+        self._submission = None
+        self._responses_cache = None
+
+    @property
+    def submission(self) -> UserFormSubmission | None:
```
EdoStorm96, line 42: When would this return None?

backend/form/services/form_evaluator.py
```diff
@@ -0,0 +1,218 @@
...
+                )
+
+        return self._submission
+
+    @property
+    def responses(self) -> dict[int, list[QuestionResponse]]:
+        """Cache all responses for this submission, grouped by question_id."""
+        if self._responses_cache is None:
+            self._responses_cache = {}
+            if self.submission:
```
EdoStorm96, line 71: `self.submission` will always be truthy right? Related to my question on R41 of this file ...

backend/form/conftest.py
EdoStorm96, line 1: I gathered from Tijmen's review of #100, that fixture names should not start with `test_`. In the case of this PR, this actually solved an issue I was experiencing with fixtures.

backend/form/services/form_evaluator.py
```diff
@@ -0,0 +1,218 @@
...
+        return self._responses_cache
+
+    def check_trigger_value(self, answer: dict, trigger_value: dict) -> bool:
+        """Check if an answer matches the trigger condition."""
+        # If trigger_value is {}, the condition is always considered met.
+        if not trigger_value:
+            return True
+
+        # Number comparisons
+        if "min" in trigger_value:
```
EdoStorm96: Can the strings used for this be put in a Django [TextChoices](https://docs.djangoproject.com/en/6.0/ref/models/fields/#enumeration-types) or an enum?

backend/form/services/form_evaluator.py
```diff
@@ -0,0 +1,218 @@
...
+        Check if a question should be shown based on show/hide conditions.
+
+        If a question has both show and hide conditions, show conditions are
+        evaluated first (i.e. take precedence).
+        """
+        show_conditions = QuestionCondition.objects.filter(
+            target_question=question, condition_type="show"
+        )
+        hide_conditions = QuestionCondition.objects.filter(
+            target_question=question, condition_type="hide"
```
EdoStorm96: I think the condition_type's should also be in a [TextChoices](https://docs.djangoproject.com/en/6.0/ref/models/fields/#enumeration-types). So we don't need to have the literal string here.

backend/form/services/form_evaluator.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def _check_conditions(
+        self, conditions: QuerySet[QuestionCondition] | QuerySet[StepCondition]
+    ) -> bool:
+        """Helper to evaluate whether any in a set of conditions are met."""
+        for condition in conditions:
+            trigger_responses = self.responses.get(condition.trigger_question.pk, [])
+            if trigger_responses:
+                if self.check_trigger_value(
+                    trigger_responses[0].answer, condition.trigger_value
```
EdoStorm96: Will `trigger_responses[0]` be the latest answer? Or the oldest? How are responses ordered?

backend/form/services/form_evaluator_test.py
EdoStorm96, line 1: I've just skimmed this. It looks good, and they pass, so I assume it is fine.

backend/form/services/user_form_resolver.py
EdoStorm96, line 1: Cool!

backend/form/services/user_form_resolver_test.py
EdoStorm96, line 1: Again, I've just skimmed this.

frontend/composables/useProcessForm.ts
```diff
@@ -164,6 +164,79 @@ function formatQuestionLocation(
...
+            if (
+                parsed &&
+                typeof parsed === "object" &&
+                "option_id" in parsed &&
+                typeof parsed.option_id === "string"
+            ) {
+                return parsed.option_id as ReturnType;
+            }
+            // If it's a primitive, return it directly
+            if (
```
EdoStorm96: Are you planning to make this happen? Answers will always be json right? So this should not happen?

backend/form/models.py
```diff
@@ -4,13 +4,29 @@
 
 user_model = get_user_model()
 
+TRIGGER_VALUE_HELP_TEXT = """
```
EdoStorm96: These are all the options we currently have right? Maybe you can make that a bit more explicit (here or in another docstring), because now this reads as if these are _some of_ the available options.

EdoStorm96 (approved):
Okay, Really nice work I think! I like it a lot! I have some minor comments, but granted you'll address these, I approve.

One issue which comes up in the PR for me is that there are a lot of strings floating around, which have important roles. Try to use some more [TextChoices](https://docs.djangoproject.com/en/6.0/ref/models/fields/#enumeration-types) or something for this.

I would also like it if the documentation for how the `trigger_value` should be formatted is a bit more explicit. Now, we just have the help_text, which is good, but a bit short.

I am kind off wondering if (in the future) we would be able to make the `trigger_value` be based upon the trigger_question's QuestionType ... So the if it's for instance a boolean, we could only select True or False as the Triggervalue with a radio button, but if it's number, we could fill in only min, max and exact etc ... But this is maybe a bit much for now, and your system will work perfectly fine I think, but I'd appreciate some more explicit documentation for the options.

Anyway, great work! I would urge you to try and keep PR's a bit smaller again, as I struggle to be as thorough, as I'd like with such a big PR ...

I should say, I haven't actually tested this yet ... But I trust that it is working, based on your testing and the tests. If you'd like me to actually test this, I can do that too :)

backend/form/services/form_evaluator_test.py
XanderVertegaal: Skimming is enough! :)

backend/form/conftest.py
XanderVertegaal: That makes sense. Fixtures for pytest are by default used for testing, so it's a bit double to name them `test_`. I'll rename them.

backend/form/models.py
XanderVertegaal: These are indeed all the options we have. (I was going to say 'for now', but I think this might actually all the values we will ever need and more.) I will add a section to the README that is more explicit about this.

backend/form/services/form_evaluator.py
XanderVertegaal: Great suggestion!

frontend/composables/useProcessForm.ts
XanderVertegaal: True! This was from an earlier prototype 😅

backend/form/services/form_evaluator.py
XanderVertegaal: Definitely, good idea!

backend/form/services/form_evaluator.py
XanderVertegaal: If there is no user submission yet and the `create_submission` flag is set to `False` (or not provided), `.first()` will return `None`. This happens when a user sees/opens a form for the first time without having interacted with it.

backend/form/services/form_evaluator.py
XanderVertegaal: In most cases, yes, but not always!

PR #115, opened 2025-12-22
Feature/study model

### LINKED ISSUE #62: Study objects and related data model
Define the shape of the generic study object, and consider sub-objects like VWR registrations, approvals, revisions, question/answer pairs

May be a implemented as code or as a document defining the data model

EdoStorm96:
fixes #62

Ok, so here is a very simple first version of how I am envisioning reviews, studies and "decisions" for now.

A simple overview:
### Study
The `Study` stays fairly bare bones. This mostly functions as a way of collecting related submitted forms.

### StudyFormSubmission
Submitted forms are `StudyFormSubmission` objects. This is a subclass of `UserFormSubmission` with some more attributes needed for this specific use. The `StudyFormSubmission` is the level where status and reviews happen, as opposed to the `Study`.

### StatusChange
The keep track of the status of a `StudyFormSubmission`, we have the `StatusChange` model. When the status of a `StudyFormSubmission` changes, as new `StatusChange` gets created and gets linked to the `StudyFormSubmission`. That way we can keep track of a `StudyFormSubmission`'s status progression. I envision `StatusChanges` as not being created directly by users, but as a side-effect of a diverse set of actions done by users. eg. starting a new form as a researcher, will always add a `CREATED` `StatusChange` to a `StudyFormSubmission` and the PO closing a review round, will add a `APPROVED` or `REJECTED` `StatusChange`, depending on the PO's verdict.

### ReviewRound
Than we have the `ReviewRound` object. This related to a `StudyFormSubmission` as well. There can be multiple `ReviewRound`'s if revisions are involved. A ReviewRound is more so a collection of `Review` objects, which contain the actual feedback on a submitted form. But finishing a `ReviewRound` will actually impact the status of a `StudyFormSubmission`, whereas `Review`'s just inform the final decision.

### Review
The `Review` object's content is open, because it will use a `UserFormDecision`. So we can easily change this over time. I envision `Review`'s as simply informative to the final decision, rather directly affecting the final decision.

Edit: I meant `UserFormSubmission`, not `UserFormDecision` up here ^. But we did not end up going this route for the `Review` object...

This is a rough draft of this system, but I am sure there is still some tweaks which need to be worked out. I just wanted a second opinion on this design, before I continue in this direction.


backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+    relate to a specific study.
+    """
+
+    study = models.ForeignKey(
+        Study, on_delete=models.CASCADE, related_name="form_submissions"
+    )
+
+    @property
+    def status(
+        self,
```
XanderVertegaal: Remove this comma and rerun Black to get formatting that is a bit less crazy ;)

backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+    ):
+        return self.status_changes.last().status
+
+
+########################
+# Status Change Object #
+########################
+
+
+class Statuses(models.TextChoices):
```
XanderVertegaal: `TextChoices` is a kind of enum, and I think the convention is for enum to use singular entities in their name, so I'd call this `Status`.

I could see ourselves adding more status-tracking enums in the future, so maybe `SubmissionStatus` is better?

backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+class StatusChange(models.Model):
+    """
+    Object which handles the status of a StudyFormSubmission. These are appended
+    to a StudyFormSubmission as a side effect for certain actions.
+    """
+
+    status = models.CharField(choices=Statuses.choices)
+
+    user_form_submission = models.ForeignKey(
+        "research.StudyFormSubmission",
```
XanderVertegaal: Since `StudyFormSubmission` is defined above, can't you simply do the following? That helps with the typing.

```suggestion
        StudyFormSubmission,
```

backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+        on_delete=models.CASCADE,
+        related_name="status_changes",
+    )
+
+    created_by = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+
```
XanderVertegaal: Spurious whiteline.

backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+########################
+
+
+class Statuses(models.TextChoices):
+    DRAFT = "DRA"
+    SUBMITTED = "SUB"
+    APPROVED = "APP"
+    REJECTED = "REJ"
+    REVISED = "REV"
+    COMPLETED = "COM"
```
XanderVertegaal: Isn't COMPLETED not implied by APPROVED or REJECTED? Is REVISED needed if we also have SUBMITTED?

Can we do with just DRAFT, SUBMITTED, APPROVED, REJECTED?

backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+
+class ReviewRound(models.Model):
+
+    reviewed_form = models.ForeignKey(
+        "research.StudyFormSubmission",
+        on_delete=models.CASCADE,
+        related_name="review_rounds",
+    )
+
+    is_active = models.BooleanField(default=True)
```
XanderVertegaal: Isn't the latest ReviewRound automatically the one that is active? I think you can use a `updated_at` field for this, which is also more versatile.

backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+
+    class Meta:
+        ordering = ["created_at"]
+
+    objects = ReviewRoundManager()
+
+
+class Review(models.Model):
+
+    review_form = models.ForeignKey(
```
XanderVertegaal: If `Review`s are bound by a `ReviewRound` and a `ReviewRound` is already bound to a `UserFormSubmission`, I don't think you need an extra foreign key here to explicitly bind a review to a form submission again.

backend/research/models.py
```diff
@@ -27,3 +32,132 @@ def can_be_created_by(user):
...
+    created_by = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+
+        ordering = ["created_at"]
+
+    def __str__(self) -> str:
+        return f"{Statuses(self.new_status).label}: {self.created_at.strftime('%d-%m-%Y, %H:%M')}"
```
XanderVertegaal: What is this `new_status`?

XanderVertegaal (commented):
Thanks for preparing this! I think your approach of suggesting new models in the form of a PR and adding an elaborate justification of why you think we need this architecture is a good way to move the discussion forward and make things more concrete! 👍

I am on board with your description of  StatusChange objects to track the 'history' of a submission. We will need this if we want to implement the 'timeline' we talked about.

A couple of things are still unclear to me; maybe this is because of my blissful ignorance of the complexities involved in FETC reviews :P In any case, I'd like to hear your thoughts on these. Chances are you have thought about these things already.

- Do we need `StudyFormSubmission` if we also have `UserFormSubmission`? Can't we add a field `study` to `UserFormSubmission`?

- The relationships between `ReviewRound`, `Review` and `UserFormDecision` are not clear to me. A `ReviewRound` is a collection of `Review`s, but aren't `Review`s not already bound/collected by the form submission they are attached to? Do we really need both? What would `UserFormDecision` be?

Again, maybe there are simple answers to these questions, but in any case they are good to keep the discussion going.

In the comments below are some minor stylistic things that are not relevant to the overall discussion.

backend/research/other_models/reviews.py
XanderVertegaal, line 1: Instead of using `other_models` as a directory next to a file `models.py`, I'd create a directory `models` and put `reviews.py` there. The old `models.py` can then perhaps be renamed to something like `study.py`, since these are all study-related models?

backend/research/models.py
```diff
@@ -39,125 +45,25 @@ def can_be_created_by(user):
 ##########################
 
 
-class StudyFormSubmissionManager(BaseMRManager):
+class StudyFormManager(BaseMRManager):
     def _viewable_objects(self, user: User):
         if user.is_privacy_officer or user.is_fetc_member:
-            return self.exclude(status=Statuses.DRAFT)
+            return self.exclude(status=SubmissionStatus.DRAFT)
```
XanderVertegaal: Will this work? This model does not have a field `status` anymore.

backend/research/models.py
```diff
@@ -39,125 +45,25 @@ def can_be_created_by(user):
...
         return self.filter(created_by=user)
 
     def _editable_objects(self, user: User):
         return self.filter(created_by=user)
 
 
-class StudyFormSubmission(UserFormSubmission):
+class StudyForm(models.Model):
```
XanderVertegaal: (Don't forget to rename the comment on line 44 as well. It still says Study Form Submissions.)

Will a `StudyForm` hold more than just references to a `Study` and a `UserFormSubmission`? If not, can't we replace it with a set of ForeignKeys? (You might have thought about this and are planning to add more fields in the future, so I'm just checking.)

XanderVertegaal (approved):
After addressing the (minor) changes in my comments, this can be merged as far as I am concerned. :) We might change some things in the future but this is a solid basis! <3

tijmenbaarda:
In the last paragraph you mention `UserFormDecision`, but I don't see that anywhere else. Did you mean something else here?

PR #123, opened 2026-01-20
Add FormLabel component

XanderVertegaal:
Outfactors repeated logic into a separate FormLabel component and marks required questions with a red asterisk "*".


EdoStorm96 (approved):
Looks fine!

PR #124, opened 2026-01-20
Feature/update question order in admin

XanderVertegaal:
Adds a StepAdminForm, mostly ~stolen~ adapted from DIAPP's ReportFieldAdmin, that lets you set the order of questions within a step. It is a bit clunky, but it works well enough for our purposes I think.

### DIFF
--- a/backend/form/forms.py
+++ b/backend/form/forms.py
@@ -53,7 +53,9 @@ class StepAdminForm(ModelForm):
         # Validate that the provided IDs exist.
         available_ids = [question.id for question in self.instance.questions.all()]
         invalid_ids = [
-            question_id for question_id in requested_order if question_id not in available_ids
+            question_id
+            for question_id in requested_order
+            if question_id not in available_ids
         ]
 
         if invalid_ids:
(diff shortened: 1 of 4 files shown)


backend/form/forms.py
```diff
@@ -0,0 +1,83 @@
...
+            for question_id in requested_order
+            if question_id not in available_ids
+        ]
+
+        if invalid_ids:
+            raise ValidationError(
+                f"Invalid question IDs: {', '.join(map(str, invalid_ids))}. "
+                f"Available IDs: {', '.join(map(str, available_ids))}"
+            )
+
```
EdoStorm96, line 69: You could also check, to see if all items from available_ids are present:
```suggestion
        missing_ids = set(available_ids) - set(requested_order)

        if missing_ids:
            raise ValidationError(
                f"Missing IDs: {', '.join(map(str, missing_ids))}. "
                f"Available IDs: {', '.join(map(str, available_ids))}"
            )
```

EdoStorm96 (approved):
Looks good! Nice feature! I have one suggestion, but do with that what you will. It's not too important.

backend/form/forms.py
XanderVertegaal: Good suggestion! 👌

PR #130, opened 2026-01-28
Feature/user form mutation

### LINKED ISSUE #111: Mutation for form submission
Maybe start with a simple version, then use that as a jumping off point.

EdoStorm96:
So here is (finally) the general FormMutation PR. This is still a proof-of-concept, as now it saves and refetches on every keystroke/change, but it does work.

I think the backend side of this is looking pretty good. The frontend still needs a little work as that is my obvious weakness. Luckily I had generous help from @XanderVertegaal.

There is still a little issue, where now it saves and refetches the form twice ... This is because we had to kind off overengineer the reactivity to get this to work properly so it seems we have created a Mutate > refetch > mutate > refetch situation. This can be solved later though. For now we have a working prototype.

I would suggest testing this with a fresh form also, as that was initially not working.


backend/api/graphql/schema.py
```diff
@@ -1,5 +1,6 @@
 from graphene import Schema, ObjectType
 
+from form.mutations.UpdateUserFormMutation import UpdateUserFormMutation
```
XanderVertegaal: Super nitpicky, but maybe you could move this down so we keep types, queries and mutations separate. Speaking from experience: the imports in this file get unwieldy really quickly :P

backend/form/mutations/utils/inputs.py
```diff
@@ -0,0 +1,17 @@
...
+    repeat_index = Int(required=True)
+
+
+class UserFormInput(InputObjectType):
+
+    id = ID()
+
+    form_config_id = ID()
+
+    responses = List(NonNull(ResponseInput), required=True)
```
XanderVertegaal, line 14:
```suggestion
class UserFormInput(InputObjectType):
    id = ID()
    form_config_id = ID()
    responses = List(NonNull(ResponseInput), required=True)
```

(Removes unnecessary whitespace)

backend/form/mutations/UpdateUserFormMutation.py
```diff
@@ -0,0 +1,57 @@
+from django.db import transaction
```
XanderVertegaal: Unused imports.

backend/form/mutations/UpdateUserFormMutation.py
```diff
@@ -0,0 +1,57 @@
...
+from django.utils import timezone
+from graphene import List, Mutation, ResolveInfo, String, Boolean
+from graphene_django.types import ErrorType
+
+from form.models import QuestionResponse, UserFormSubmission
+from form.mutations.utils.inputs import UserFormInput
+from form.types.UserFormType import UserFormType
+
+
+class UpdateUserFormMutation(Mutation):
```
XanderVertegaal: You're not really updating a form but a form _submission_ here, so maybe `UpdateUserFormSubmission` is slightly more accurate?

backend/form/mutations/UpdateUserFormMutation.py
```diff
@@ -0,0 +1,57 @@
...
+
+    ok = Boolean(required=True)
+    errors = List(String, required=True)
+
+    @classmethod
+    def mutate(
+        cls,
+        root: None,
+        info: ResolveInfo,
+        user_form_input,
```
XanderVertegaal: In this case the type of `user_form_input` is already clear from the context, but if you wish to document it explicitly, you can of course do so:

```suggestion
        user_form_input: UserFormInput,
```

In my IDE this gives me a red squiggly if I try to do `user_form_input["id"]`. The code works, but I get an error message in my IDE: '"__getitem__" method not defined on type "UserFormInput" (PylancereportIndexIssue)'. To avoid this error, I found that you can use [`getattr(...)`](https://www.geeksforgeeks.org/python/python-getattr-method/). You can even provide a default value if you are not sure the property exists: `getattr(user_form_input, "id", None)`.

This is not something I insist you implement here, but just meant as a general tip for future mutations with more complicated inputs ;)

backend/form/mutations/UpdateUserFormMutation.py
```diff
@@ -0,0 +1,57 @@
...
+                    qr.save()
+            except QuestionResponse.DoesNotExist:
+                print(
+                    [
+                        submission.id,
+                        response.question_id,
+                        response.repeat_index,
+                        response.id,
+                    ]
+                )
```
XanderVertegaal: Rogue print statement.

backend/form/mutations/UpdateUserFormMutation.py
```diff
@@ -0,0 +1,57 @@
...
+from form.mutations.utils.inputs import UserFormInput
+from form.types.UserFormType import UserFormType
+
+
+class UpdateUserFormMutation(Mutation):
+    class Arguments:
+        user_form_input = UserFormInput(required=True)
+
+    ok = Boolean(required=True)
+    errors = List(String, required=True)
```
XanderVertegaal:
```suggestion
    errors = List(NonNull(ErrorType), required=True)
```

backend/form/mutations/UpdateUserFormMutation.py
```diff
@@ -0,0 +1,57 @@
...
+                    ]
+                )
+                QuestionResponse.objects.create(
+                    submission=submission,
+                    question_id=response.question_id,
+                    answer=response.answer,
+                    repeat_index=response.repeat_index,
+                )
+            except:
+                error = ErrorType(messages=["something went wrong ..."])
```
XanderVertegaal: This error message could be a bit more specific/helpful.

backend/form/services/form_evaluator.py
```diff
@@ -221,12 +221,12 @@ def is_step_visible(self, step: Step) -> bool:
 
         return True
 
-    def get_user_answer(
+    def get_user_response(
         self, question: BaseQuestion, repeat_index: int = 0
     ) -> dict | None:
```
XanderVertegaal:
```suggestion
    ) -> QuestionResponse | None:
```

backend/form/services/form_evaluator_test.py
```diff
@@ -658,8 +658,10 @@ def test_get_user_answer(self, form, test_user, trigger_question):
 
         evaluator = FormEvaluator(form, test_user)
 
-        assert evaluator.get_user_answer(trigger_question, 0) == {"value": "my answer"}
-        assert evaluator.get_user_answer(trigger_question, 1) is None
+        assert evaluator.get_user_response(trigger_question, 0).answer == {
```
XanderVertegaal: You can fix the type warnings by adding an assertion (and fixing the type signature in `form_evaluator.py`, q.v.).


```suggestion
        response = evaluator.get_user_response(trigger_question, 0)
        assert response is not None 
        assert response.answer == {
```

frontend/components/shared/FormWrapper.vue
```diff
@@ -55,6 +138,20 @@ function getAllSteps(form: FormWithValues): CombinedStepWithValues[] {
     });
 }
 
+function getAllQuestions(
```
XanderVertegaal, line 145: Nice!

frontend/components/shared/FormWrapper.vue
```diff
@@ -55,6 +138,20 @@ function getAllSteps(form: FormWithValues): CombinedStepWithValues[] {
...
+    step: StepWithValues | SubstepWithValues,
+): QuestionWithValue[] {
+    // helper function to get all questions as a flat list
```
XanderVertegaal: General tip: if you put these function-wide comments in the docstring above, they become visible when you mouse over the function name elsewhere:

```suggestion
/**
 * Returns all questions from a step in a flat list.
 */
function getAllQuestions(
    step: StepWithValues | SubstepWithValues,
): QuestionWithValue[] {
```

frontend/components/shared/FormWrapper.vue
```diff
@@ -7,14 +7,97 @@ import MRForm from "./MRForm.vue";
...
+            })
+            .catch((error: unknown) => {
+                console.error("Error updating form:", error);
+            });
+    }
+}
+
+function FormDataToMutationInput(
+    formData: FormWithValues,
+): UserFormInput {
```
XanderVertegaal: Camelcasing is the norm for JS/TS functions.

```suggestion
function formDataToMutationInput(formData: FormWithValues): UserFormInput {
```

frontend/components/shared/FormWrapper.vue
```diff
@@ -7,14 +7,97 @@ import MRForm from "./MRForm.vue";
...
+    // Utility function to transform our form into the expected input for our mutation
```
XanderVertegaal: This comment is helpful, but I'd put it in a docstring (see comment at `getAllQuestions` below).

The other two comments in this function are a bit superfluous, IMO.

frontend/components/shared/FormWrapper.vue
```diff
@@ -7,14 +7,97 @@ import MRForm from "./MRForm.vue";
...
+
+    // First collect all questions
+    const questions = formData.steps.flatMap(getAllQuestions);
+
+    // Then return out object in its expected form
+    return {
+        id: props.queriedForm.submissionId ?? null,
+        formConfigId: props.queriedForm.formId,
+        responses: questions.map(
+            (question: QuestionWithValue): ResponseInput => {
```
XanderVertegaal, line 91: You shouldn't have to type the input and output here explicitly. TS can infer them from the function signature. (It doesn't hurt though.)

frontend/pages/procreg/[slug].vue
```diff
@@ -128,6 +131,7 @@ function stepSlug(route: string | string[]): string {
                 v-if="form"
                 :queried-form="form"
                 :current-step-slug="stepSlug(route.params.slug)"
+                :refetch-form="refetch"
```
Comment (XanderVertegaal) [inline frontend/pages/procreg/[slug].vue]: Instead of passing the `refetch` function from the parent to the child, I wonder if it makes more sense to send ("emit") a signal up from the child to the parent and have the parent call `refetch()`. That way, both the fetching and refetching are done in the same component (`[slug].vue`).

This is again a minor / stylistic nitpick. Feel free to ignore it if you'd like.

backend/form/mutations/UpdateUserFormMutation.py
XanderVertegaal: Changing this will force you to update your query in the frontend as well, but it is more accurate.

backend/form/mutations/utils/inputs.py
```diff
@@ -0,0 +1,17 @@
...
+class ResponseInput(InputObjectType):
+    id = ID()
+    question_id = ID(required=True)
+    answer = JSONString(required=True)
+    repeat_index = Int(required=True)
+
+
+class UserFormInput(InputObjectType):
+
+    id = ID()
```
XanderVertegaal: Would `submission_id` be more accurate/clear?

XanderVertegaal (approved):
Nice work! 🎉 The code changes are kept nice and simple, and functionally it works perfectly (as far as I can tell). The comments here are mostly general tips / stylistic. 

The only thing I think you should change is the type of `UpdateUserFormMutation.errors`, which is currently inaccurate. Everything else is optional.

PR #137, opened 2026-01-29
Add notifications component and use it when list fetching fails

tijmenbaarda:
The notification will show any time a GraphQL list fails to load.

To test it, make list fetching fail in the backend, for example by adding a line that does not work in `resolve_study_pages()`.


EdoStorm96 (changes requested):
So this is working nicely! Good work. 

I think its fine, although I would prefer to make use of [bootstrap's toast](https://getbootstrap.com/docs/5.3/components/toasts/) class. Then we don't have to reinvent this complicated CSS. Do you have time to implement this? Or would you prefer if someone else picks it up?

XanderVertegaal:
This PR is subsumed in #153

PR #139, opened 2026-02-02
Add debounce and dedupe requests

XanderVertegaal:
This is an addition to #111, which should be merged first. It makes sure submissions are debounced with 300ms (we can adjust this of course). It also adds a small mechanism that compares mutation inputs and avoids duplicate mutations.

There is still the issue that changes made between mutations and refetches are sometimes lost: you can test this by typing in a text field: some characters are 'eaten'/removed. This issue will become less problematic once we implement more sophisticated logic to determine when a mutation should be sent (e.g. by using an 'onBlur' event, or having the user click a 'Submit' or 'Next Step' button).

### DIFF
--- a/backend/conftest.py
+++ b/backend/conftest.py
@@ -2,7 +2,6 @@ import pytest
 
 from django.contrib.auth import get_user_model
 
-
 User = get_user_model()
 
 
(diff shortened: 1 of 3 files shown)


XanderVertegaal:
The relevant bits from this PR will be subsumed in #144.

PR #140, opened 2026-02-04
feat: docker settings to allow for debugging + django-extensions for shell_plus

EdoStorm96:
I know both @miggol and me are quite fond of using breakpoints and using pdb for debugging. This wasn't possible, but using these settings for the dev container does allow for this when attaching to the django container. It is still a bit annoying that the healthcheck can interfere ... But this is an improvement for now.


EdoStorm96:
I've also just pushed a commit to add django-extensions. This add the shell_plus command, which I find really useful for interacting with the DB. I also use this a lot on production for ethics, which is why I've added it to generic_settings.py

miggol (approved):
Big fan!

PR #144, opened 2026-02-18
Feature/submission timing

XanderVertegaal:
(This replaces #139.)

Features:
- Submissions are triggered more intelligently, i.e. only when 
  - the user switches steps (either by pressing the 'Previous'/'Next' step buttons or interacting with the stepper, or
  - a question with associated conditions has changed
- For TextQuestions and NumberQuestions, a small debounce of 300 ms is added to decrease the risk of inputs being 'eaten'.

Bonus:
- `formDataToMutationInput` has been outfactored into its own composable to make `FormWrapper.vue` a bit leaner.
- The queried form is now evicted from Apollo's cache after every submission. This automatically triggers a refetch of the form, so we don't have to emit a `formSaved` event and manually refetch anymore.


backend/form/models.py
```diff
@@ -270,13 +282,6 @@ class StepCondition(BaseCondition):
         help_text="The step that this condition applies to.",
     )
 
-    trigger_question = models.ForeignKey(
```
XanderVertegaal, line 273: Deleting `trigger_question` here (and in `QuestionCondition`) ensures that these models make use of `trigger_question` in the base class `BaseCondition` instead.

frontend/components/shared/FormWrapper.vue
```diff
@@ -52,52 +33,40 @@ const UPDATE_USER_FORM = graphql(`
                 responses: $responses
             }
         ) {
+            ok
```
XanderVertegaal, line 36: I put simple fields (string/number/ID) above complex fields (objects/arrays) to keep queries readable. This is personal preference more than anything else,

frontend/components/shared/MRForm.vue
```diff
@@ -1,14 +1,35 @@
-<script lang="ts" setup>
-import {
-    SharedDateQuestion,
-    SharedFileUploadQuestion,
-    SharedNumberQuestion,
-    SharedSelectQuestion,
-    SharedTextQuestion,
-    SharedTrueFalseQuestion,
-} from "#components";
+<script lang="ts">
```
XanderVertegaal, line 1: These script tags without `setup` are only interpreted once, not every time the component is mounted or unmounted. This makes them perfect for constants.

frontend/components/shared/FormWrapper.vue
```diff
@@ -52,52 +33,40 @@ const UPDATE_USER_FORM = graphql(`
...
 `);
 
-const { mutate: mutateForm } =
-    useMutation<UpdateUserFormSubmission>(UPDATE_USER_FORM);
+const { mutate: mutateForm } = useMutation<UpdateUserFormSubmission>(
+    UPDATE_USER_FORM,
+    {
+        update: (cache) => {
+            cache.evict({ fieldName: "form" });
+            cache.gc();
```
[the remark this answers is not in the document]
XanderVertegaal: We're only going to use one form as of now.

backend/form/models.py
```diff
@@ -216,6 +227,7 @@ class TriggerValueKeys(models.TextChoices):
         BaseQuestion,
         on_delete=models.CASCADE,
         help_text="The question whose answer triggers this condition.",
+        related_name="triggered_%(class)s",
```
EdoStorm96, line 230: Wow, this is pretty crazy with the formatting. I do find it a little confusing though and am wondering when we'd need to differentiate between the two types? BaseQuestion.has_conditions could for instance be simplified if we'd get rid of the formatting. But maybe I am overlooking something?

frontend/components/shared/FormStepper.vue
EdoStorm96, line 1: The centering of the stepper got thrown off. Not a huge deal, as we'll need to look at the CSS for this some more anyway, but if you can fix it, would be nice.

EdoStorm96 (approved):
All looks great. I found some tiny tiny things, but in general this is working great and I really like how you've implemented it! You can decide if you want to fix the stuff I found, they're really nitpicks.

backend/form/models.py
XanderVertegaal: I know! It's quite neat that Django allows this. This was chiefly to maintain the old situation, in which we already differentiated step conditions from question conditions. It doesn't hurt, I think, and it is a bit more explicit/concrete than relying on the base class.

frontend/components/shared/FormStepper.vue
XanderVertegaal: I'd be happy to pick this up, but the stepper looks the same as it does on `develop` for me.

(left: my branch, right: develop)
image

Could you check to see if it's an issue with this branch or with develop? (It looks good to me, but you are more familiar with this component than I am, because of Ethics ;) )

EdoStorm96: Screenshot from 2026-02-26 09-41-41
Screenshot from 2026-02-26 09-37-46

Mmmm, for me this is the difference between this branch and develop ... Not sure what causes this, but we can look at it later.

XanderVertegaal: I had to look twice, but indeed the step 'balls' are not aligned anymore. Weird! I'll make a separate issue for this so this can be merged.

XanderVertegaal: #160

PR #147, opened 2026-02-18
Feature/automatic study creation

### LINKED ISSUE #112: Study creation mechanism
This feels like it should be a different mutation than simply "saving" a formconfig halfway through. Submitting the roadmap is the final step in the form, which then creates the study.

At least, this is how we're imagining this right now.

XanderVertegaal:
Closes #112 

This PR sets up the groundwork for the automatic creation of a new Study as part of a user form submission.

The `UserFormInput` now has an optional `create_study` argument. If set to true, this will create a study as a side effect. As the name of study will be equal to the value of the question referred to using the newly added `MRForm.study_name_question` field. Since the name of a Study is now computed based on this field, the existing `Study.title` field is removed.

The frontend currently does not include the `create_study` flag. To test the creation of a study, modify UPDATE_USER_FORM in FormWrapper.vue to include `createStudy: true`

```ts
const UPDATE_USER_FORM = graphql(`
    mutation SaveFormSubmission(
        $submissionId: ID
        $formConfigId: ID
        $responses: [ResponseInput!]!
    ) {
        updateFormSubmission(
            userFormInput: {
                submissionId: $submissionId
                formConfigId: $formConfigId
                responses: $responses
                createStudy: true  // <-- Add this
            }
        ) {
...
```

Alternatively, use the following test mutation on `.../api/graphql`:

```graphql
mutation SaveFormSubmission {
  updateFormSubmission(
    userFormInput: {
      submissionId: "1", 
      formConfigId: "1", 
      responses: [], 
      createStudy: true
    }
  ) {
    ok
    errors {
      field
      messages
    }
  }
}
```


backend/research/models.py
```diff
@@ -21,17 +16,43 @@ def _editable_objects(self, user: User):
...
+        default_name = (
+            f"Study created on {self.created_at.strftime('%Y-%m-%d %H:%M:%S')}"
+        )
+
+        name_question = self.form.study_name_question
+        if name_question is None:
+            return default_name
+
+        try:
+            submission = UserFormSubmission.objects.get(
```
EdoStorm96, line 43: This could return multiple submissions, if revisions are involved. I think it might be useful to store the name on the Study model, once we have it, so then this function could just resolve `self.name` We would also want the study name to be permanent I guess?

backend/form/models.py
```diff
@@ -7,10 +7,20 @@
 
 class MRForm(models.Model):
     name = models.CharField(max_length=200)
+    version = models.CharField(max_length=10, help_text="E.g. 1.0.0, 2.15.3")
 
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    study_name_question = models.ForeignKey(
```
EdoStorm96, line 15: Mmm ... I'm not sure if I really like this living here ... I see the purpose of having the name be a "special" question, but I also think it is quite confusing. Can't we just use a normal question object to ask for the Study name? We'd maybe need a constant or something to keep track of important questions like this ... but it does save use from having this field here. If we were to have it as an attribute on a model, to me it would make more sense on the UserFormSubmission.

backend/form/mutations/utils/inputs.py
```diff
@@ -12,3 +12,4 @@ class UserFormInput(InputObjectType):
     submission_id = ID()
     form_config_id = ID()
     responses = List(NonNull(ResponseInput), required=True)
+    create_study = Boolean(required=False, default_value=False)
```
EdoStorm96, line 15: I'm not sure if I like this design ... I think it could be possible we have more complex conditions for creating a study (like AND and OR logic and stuff ...) and it would be more flexible to evaluate responses using custom logic, as long as there is no study connected to a UserFormSubmission.

backend/research/models.py
```diff
@@ -21,17 +16,43 @@ def _editable_objects(self, user: User):
 
 
 class Study(models.Model):
+    form = models.ForeignKey(MRForm, on_delete=models.PROTECT)
```
EdoStorm96: We could get to the form via the submissions connected to a study, so this feels unnecessary.

EdoStorm96 (changes requested):
I think this is a good start, but I don't fully love this design yet ... I think that is partly because I have the revision PR in mind, But I feel that this is not the simplest or most flexible way to go about this, and that we should try to stick closer to using our form system as it is designed, for deciding when to make a study and to ask for a name.

I can also could imagine that the logic for when to create a study could get more complex over time, so I'd prefer to have more flexibility in creating this logic vs. just having a flag on one question.

I think it would be best if the revisions branch is merged before this PR as it will cause some merge issues. I'll get to work on that!

Let me know what you think! We can discuss it further on wednesday.

backend/form/mutations/UpdateUserFormSubmission.py
```diff
@@ -19,34 +21,64 @@ def mutate(
...
+            return cls(ok=False, errors=[error])  # type: ignore
+
+        if create_study_flag:
+            create_study(submission)
+
+        return cls(ok=True, errors=[])  # type: ignore
+
+
+def create_study(submission: UserFormSubmission) -> None:
+    study, created = Study.objects.get_or_create(
```
EdoStorm96: It does not really make sense to use get_or_create here to me, because a user could have multiple studies going with the same form here right? I think it would be safe to assume that `if create_study_flag`, a new study should be created. If you are really set on trying to deal with the possibility we are sending `create_study_flag`, when we shouldn't, this get_or_create should have more specific arguments, but I say lets just try and ensure we only try to send `create_study: true` to the backend, when we should.

backend/form/mutations/UpdateUserFormSubmission.py
EdoStorm96, line 1: Yeah, so this file will cause some merge issues with the revisions PR, but we can handle that when we get to it.

backend/research/models.py
EdoStorm96: Ok, so maybe the study name does not have to be permanent actually as I'm thinking more about it. We can use the ref number for that. And then this method should work fine :) I will try to implement Michael's annotate idea, in my revisions branch, so then you could easily get the latest submission too ...

EdoStorm96 (commented):
Ok, so I am really sorry about my sloppy first review. On second review, this is quite good actually. I just would like for you to adress my comments about the get_or_create study method, but then this is pretty much good to go, as far as I'm concerned. We have to decide if we are gonna merge the revisions branch first or this one, as they will cause some merge issues. My preference would be to merge revisions first if that is ok with you?

backend/form/mutations/UpdateUserFormSubmission.py
XanderVertegaal: No worries, I'm on it :)

backend/form/mutations/UpdateUserFormSubmission.py
XanderVertegaal: Ah, it wasn't clear to me that a user could have multiple studies with the same form, but it makes sense. I'll adapt this!

backend/research/models.py
XanderVertegaal: I'll turn it into a property.

XanderVertegaal:
For your re-reviewing pleasure, here is a list of changes since your last review:

- Marked `UserFormInput.form_config_id` as required. The frontend already always provides a value for this.
- `UpdateUserFormSubmission`: let error in `save_responses()` bubble up to `mutate()`, where it is handled.
- Added related_name `"submissions"` to: 
	- `UserFormSubmission.form` (used to be none)
	- `UserFormSubmission.user` (used to be none)
	- `UserFormSubmission.study` (used to be `"forms"`, which looks like a mistake)
- Refactored model field `Study.form` into a computed property, which returns the form that is linked to the study's submissions. This should always be one and the same form. If there is none or more than one, it will raise an exception (on purpose).

backend/form/mutations/UpdateUserFormSubmission.py
```diff
@@ -34,4 +41,45 @@ def mutate(
...
             return cls(ok=False, errors=[error])
 
-        return cls(ok=True, errors=[])
+        try:
+            save_responses(submission.pk, responses)
+        except Exception as e:
+            error = ErrorType(
+                field="responses",
+                messages=[
+                    "Failed to save responses for UserFormSubmission with id: {submission_id}"
```
EdoStorm96: This should be an f-string right?

backend/form/mutations/UpdateUserFormSubmission.py
EdoStorm96: This comment is still not adressed. I think we should just always create a study here.

backend/form/mutations/UpdateUserFormSubmission.py
EdoStorm96: Ok, so I know this is a bit annoying, but in the revision pr I refactored this mutation to make use of a service (`form.services.update_submission`). I think this is the way to go. Could you change this branch to also make use of this service?

backend/research/models.py
```diff
@@ -21,17 +16,47 @@ def _editable_objects(self, user: User):
...
         return user.is_authenticated
 
+    @property
+    def form(self) -> MRForm:
+        # There should only ever be one form associated with a study.
+        # If there are none or more than one, we will want to know about it.
+        return MRForm.objects.filter(submissions__study=self).distinct().get()
+
+    @property
+    def name(self) -> str:
```
EdoStorm96, line 33: I am still having some doubts about this method. It will be pretty slow if we try to get a bunch of studies and therefore I feel like it would also be logical to store it in the db, once we have a value ... Let's discuss it in our dev meeting.

EdoStorm96 (changes requested):
This is an improvement, but I still have some of the same issues. If you want me to elaborate, give me a call :)

miggol (changes requested):
Could you merge develop into this, and update `create_dev_data.py` so it works again? I notice you might have to rearrange a migration, but the sql scripts are now in develop so that shouldn't be too bad.

Right now `create_dev_data` is my only blocker. I'm looking into getting the study name as an annotation in a custom manager, but it's hard to debug and test so far behind develop without dev data.

I don't consider the study name issue a blocker, as long as Edo is okay with it for now (and `create_dev_data` works) I'm fine with merging this as-is to keep things moving along. We can treat the study name thing as a separate issue.

I can't speak for Edo's other comments though, I'm a bit out of the loop with the `update_submission` service.

XanderVertegaal:
> Right now `create_dev_data` is my only blocker. I'm looking into getting the study name as an annotation in a custom manager, but it's hard to debug and test so far behind develop without dev data. 

`create_dev_data` should now work again. (No migration changes needed; just a different way to create Study objects.) See if it works for you as well now.

miggol (approved):
Let's do #203 separately and merge this now. Thanks for all the adjustments

PR #152, opened 2026-02-23
Feature/revisions

### LINKED ISSUE #136: Revisions
How do they work? How do we keep answers from previous revisions intact? Here's a silly drawing:

Image

https://excalidraw.com/#json=HhgEu23zBlY_DhH2L7URk,G4P3M5TWoXfW-1gS8F_uGQ

Revisions in this drawing can be seen as empty objects, that simply exist to link answers. For this reason, they are very easy to copy.

EdoStorm96:
fixes #136 

Ok, so here is a backend implementation of revisions. This took a bit longer than expected, so I thought it would be good to ask for a review now. Also, given our new prioritization with the milestones, I think we'll leave implementing the rest of revisions for later. But this is a good starting point anyway.

As discussed in the issue, a revision is basically a copy of an old submission with a new pk. It still points to the same responses. Then once a new answer is given to a question, it removes the old response from the revision and creates a new one.

What mainly caused a headache when implementing this is the fact that the relationship between responses and submissions is now a M2M and not a foreignkey. I still believe that it would be useful to also add a foreignkey to the first submission to which a response was added, but since you were opposed to this I tried to avoid this.  We could override `save()` and have this set automatically I think ... 

Some (not super important) features that we did have previously, I could not get to work anymore. These are:

- Some filters and search fields in `QuestionResponse`'s admin page
- `unique_together = ["submission", "question", "repeat_index"]` on `QuestionResponse`

If anyone has tips for getting the unique constraint to work again, I'd like to hear them. I think having a separate ForeignKey to the first submission would also be helpful for this.

I do have a wip stash of trying to implement custom filters to get back some of the functionality to the admin page ... but ultimately I was spending way too much time on this, and I don't think these filters and searches are that important, as I think we will be going to responses via a submission, and not going to search for responses while searching for a user.

Otherwise, this is all working fine now! The mutation works and the tests pass. Once we get this done, I will start working on things that get us closer to our nearest milestone and leave this be for a bit.


backend/form/mutations/CreateUserFormRevision.py
```diff
@@ -0,0 +1,31 @@
+from graphene import List, Mutation, NonNull, ResolveInfo, String, Boolean, ID
+from graphene_django.types import ErrorType
+
+from form.services.create_user_form_revision import create_user_form_revision
+from form.models import QuestionResponse, UserFormSubmission
```
XanderVertegaal: Unused imports.

backend/form/mutations/CreateUserFormRevision.py
```diff
@@ -0,0 +1,31 @@
...
+from form.mutations.utils.inputs import UserFormInput
+
+
+class CreateUserFormRevision(Mutation):
```
XanderVertegaal, line 7: Not a comment/request for change, but just a thought: out of habit I always name mutations `...Mutation`, but the official GraphQL documentation does not promote any naming scheme and uses mutations like `CreateReview`, which I like better than appending `Mutation` to every mutation name. I now want to change the names of the Create/Update/DeleteStudyMutation too :P

backend/form/mutations/CreateUserFormRevision.py
```diff
@@ -0,0 +1,31 @@
...
+    class Arguments:
+        submission_id = NonNull(ID)
+
+    ok = Boolean(required=True)
+    errors = List(NonNull(ErrorType), required=True)
```
XanderVertegaal, line 12: I think class fields (i.c. `ok` and `errors`) are usually put before the `Arguments` child class.

backend/form/mutations/CreateUserFormRevision.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @classmethod
+    def mutate(
+        cls,
+        root: None,
+        info: ResolveInfo,
+        submission_id: ID,
```
XanderVertegaal: `graphene.ID` is mapped to a `str` in Python. You'll also have to adjust the type signature of `create_user_form_revision()`.

```suggestion
        submission_id: str,
```

backend/form/mutations/CreateUserFormRevision.py
```diff
@@ -0,0 +1,31 @@
...
+    ):
+        try:
+            create_user_form_revision(submission_id)
+        except:
```
XanderVertegaal: I'd prefer catching individual exceptions instead of a rather broad catch-all 'Something went wrong'. Future us will thank you for it when we get errors on Sentry and don't know whether the exception was raised from `.get` or `.create`.

backend/form/models.py
```diff
@@ -173,9 +173,17 @@ def __str__(self) -> str:
 class QuestionResponse(models.Model):
     """Stores a user's answer to a question."""
 
-    submission = models.ForeignKey(
-        UserFormSubmission, on_delete=models.CASCADE, related_name="responses"
-    )
+    submissions = models.ManyToManyField(UserFormSubmission, related_name="responses")
+
+    @property
```
XanderVertegaal: Computed properties are usually put after regular model fields, I think.

backend/form/conftest.py
```diff
@@ -1,4 +1,6 @@
 import pytest
+from form.mutations.utils.inputs import ResponseInput, UserFormInput
+from main.models import User
```
XanderVertegaal: Unused import

backend/form/mutations/UpdateUserFormSubmission.py
```diff
@@ -1,6 +1,7 @@
 from graphene import List, Mutation, NonNull, ResolveInfo, String, Boolean
 from graphene_django.types import ErrorType
 
+from form.services.update_submission import update_or_create_submission
 from form.models import QuestionResponse, UserFormSubmission
```
XanderVertegaal: Unused imports

backend/form/services/update_submission.py
```diff
@@ -0,0 +1,46 @@
...
+from form.models import UserFormSubmission, QuestionResponse
+from form.mutations.utils.inputs import UserFormInput
+
+
+def update_or_create_submission(user: User, user_form_input: UserFormInput):
+    try:
+        current_submission = UserFormSubmission.objects.get(
+            id=user_form_input["submission_id"]
+        )
+    except Exception as e:
```
XanderVertegaal:
```suggestion
    except UserFormSubmission.DoesNotExist:
```

backend/form/services/update_submission.py
```diff
@@ -0,0 +1,46 @@
...
+                        question_id=response["question_id"],
+                        answer=response["answer"],
+                        repeat_index=response["repeat_index"],
+                    )
+                    new_response.submissions.add(current_submission)
+                else:
+                    # If this is not a revision, just update the answer
+                    qr.answer = response["answer"]
+                    qr.save()
+        except (QuestionResponse.DoesNotExist, KeyError) as e:
```
XanderVertegaal: I guess the `KeyError` is used for handling cases in which the response does not have an ID, but it's a bit implicit. Would it be possible to do something like

```py
for response in user_form_input["responses"]:
    response_id = getattr(response, "id", None)
```

and then base the rest on whether `response_id` is `None` or not?

backend/form/services/update_submission.py
```diff
@@ -0,0 +1,46 @@
...
+            new_reponse = QuestionResponse.objects.create(
```
XanderVertegaal: 🥐

```suggestion
            new_response = QuestionResponse.objects.create(
```

backend/form/services/create_user_form_revision.py
```diff
@@ -0,0 +1,17 @@
+from form.models import UserFormSubmission
+
+
+def create_user_form_revision(submission_id: int):
```
XanderVertegaal:
```suggestion
def create_user_form_revision(submission_id: str) -> UserFormSubmission:
```

backend/form/services/create_user_form_revision_test.py
```diff
@@ -0,0 +1,69 @@
...
+        submission_revision = create_user_form_revision(submission.id)
+
+        assert submission_revision.user == test_user
+        assert submission_revision != submission
+        assert set(submission.responses.all()) == set(
+            submission_revision.responses.all()
+        )
+
+    def test_update_revision_response(self, test_user, user_form_input):
+        """Test updating an aswer for revision"""
```
XanderVertegaal:
```suggestion
        """Test updating an answer for revision"""
```

backend/form/services/create_user_form_revision_test.py
```diff
@@ -0,0 +1,69 @@
...
+from form.services.create_user_form_revision import create_user_form_revision
+from form.services.update_submission import update_or_create_submission
+from form.mutations.utils.inputs import UserFormInput
+from form.models import UserFormSubmission, QuestionResponse
+
+
+@pytest.mark.django_db
+class TestCreateRevision:
+    """
+    Tests for create_user_from_revision service
```
XanderVertegaal:
```suggestion
    Tests for create_user_form_revision service
```

backend/form/services/create_user_form_revision_test.py
```diff
@@ -0,0 +1,69 @@
...
+        assert updated_revision == submission_revision
+        assert set(updated_revision.responses.all()) == set(
+            submission_revision.responses.all()
+        )
+
+        # If we add a new answer, a new response should get made
+        new_ans = {"value": "new_answer"}
+        user_form_input["responses"][0]["answer"] = new_ans
+
+        old_ans = old_response.answer
```
XanderVertegaal: Unused variable. Perhaps to be used on line 68, although `old_response.answer` works well enough?

backend/form/services/create_user_form_revision_test.py
```diff
@@ -0,0 +1,69 @@
...
+        # response should remain intact.
+
+        updated_revision = update_or_create_submission(test_user, user_form_input)
+
+        assert updated_revision == submission_revision
+        assert set(updated_revision.responses.all()) == set(
+            submission_revision.responses.all()
+        )
+
+        # If we add a new answer, a new response should get made
```
XanderVertegaal, line 65: I'd split this test into two at this point, to test the creation of a new answer and the update of an older answer separately.

backend/form/services/create_user_form_revision.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Util function to copy a UserFormSubmission, used for revisions.
```
XanderVertegaal, line 6: Could you expand the documentation a bit? I understand how we envisage revisions to work because we talked about it recently, but new developers and future me will likely be confused if there is a function that claims to create a revision while actually it is copying a submission.

Maybe an addition to the README would be a nice touch as well, but I'll leave that up to you.

XanderVertegaal (approved):
Nice work! 👌 I especially like the tests and the outfactoring of mutation logic into smaller services/utils. As always I have a number of smaller comments, but nothing fatal/urgent (mainly typos).

I appreciate that you kept `first_submission` a computed property ;) Let's see how workable that is in practice.

backend/form/migrations/0005_alter_questionresponse_unique_together_and_more.py
```diff
@@ -0,0 +1,28 @@
...
+        migrations.AddField(
+            model_name="questionresponse",
+            name="submissions",
+            field=models.ManyToManyField(
+                related_name="responses", to="form.userformsubmission"
+            ),
+        ),
+        migrations.RemoveField(
+            model_name="questionresponse",
+            name="submission",
```
XanderVertegaal, line 26: If you have time to spare (which is unlikely) you could consider making this migration reversible, but that's not necessary.

backend/form/models.py
```diff
@@ -173,9 +173,8 @@ def __str__(self) -> str:
 class QuestionResponse(models.Model):
     """Stores a user's answer to a question."""
 
-    submission = models.ForeignKey(
-        UserFormSubmission, on_delete=models.CASCADE, related_name="responses"
-    )
+    submissions = models.ManyToManyField(UserFormSubmission, related_name="responses")
+
     question = models.ForeignKey(BaseQuestion, on_delete=models.CASCADE)
```
miggol, line 202: As long as a question has answers, `on_delete` should probably be `models.PROTECT`

backend/form/models.py
```diff
@@ -173,9 +173,8 @@ def __str__(self) -> str:
 class QuestionResponse(models.Model):
```
miggol, line 195: A quick example of how `first_submission` could be achieved with a custom manager:

```python3
class ResponseManager(models.Manager):

    def get_queryset(self):
        base = super().get_queryset()
        return base.annotate(first_submission_pk=Min("submissions__pk"))

    def for_submission(self, submission):
        base = self.get_queryset().filter(
            submissions__pk=submission.pk)
        return base.annotate(is_new=Case(
            When(
                first_submission_pk=submission.pk, then=True,
            ),
            default=False,
        ))                
                            

class QuestionResponse(models.Model):
    """Stores a user's answer to a question."""

    objects = ResponseManager()

    # ... etc
```

This will provide the first submission on every item in `QuestionResponse.objects`.

I have also provided an example of how to annotate an `is_new` boolean when fetching all responses for a given submission using `QuestionResponse.objects.for_submission(submission)`.

Though admittedly you would prefer to do that last part on the RelatedManager on the UserFormSubmission model. I just put it here for the example, the annotate call is the interesting part.

miggol: Totally fine to keep the property as well, as long as we try to keep things in the queryset whenever there's a risk of N+1

PR #153, opened 2026-02-24
Feature/notifications

XanderVertegaal:
This PR is meant as a rework of #137 but is not very different from it in the end. It turns out that BS only offers a couple of basic styling classes and nothing else. Keeping track of notifications in a store and creating/dismissing them still has to be managed by us. Also the extra CSS which shows the user for how long the notification is active is a nice feature to keep in, I think.

Differences:
- No dependency on DayJS, no additional plugins; we now only use native date methods.
- More modern ('setup') syntax for the NotificationStore.
- Simpler notification templates: now they always have a body and a header, with icons.
- Fixed animations, using Vue's built-in `TransitionGroup`: inserts and removes list items with a little animation.


EdoStorm96 (approved):
Yes, this looks great! Nice job streamlining this a little.

PR #163, opened 2026-03-04
Postgres housekeeping scripts

miggol:
These scripts allow you to dump, wipe, and restore your postgres DB.

I was missing these features from the SQLite days, mostly for debugging migrations and easily moving data between worktrees.


scripts/pg_wipe.sh
```diff
@@ -0,0 +1,21 @@
...
+echo This will destroy your current database. You may want to run pg_dump.sh to make a backup first.
+echo Continue? [y/N]
+read y
+
+if [[ $y != y && $y != Y ]]; then
+    echo Aborting
+    exit 1
+fi
+
+if docker compose --profile dev run mr-postgres sh -c \
```
XanderVertegaal, line 18: You might not need the `--profile dev`. Mr postgres runs in both profiles.

XanderVertegaal (commented):
I'm still running MR locally so I'm not really in a position to check these scripts functionally. Code-wise they seem good. Nice work! 👍 

I appreciate that you added a README! Perhaps you can add that these scripts are not meant for prod, or find a way to restrict them to dev builds.

Maybe with something like:

```sh
if docker compose ps --profile prod --status running | grep -q "mr-"; then
    echo "ERROR: Production containers are running. These scripts are for development only."
    exit 1
fi
```

Or do you intend these commands to be run on prod as well?

miggol:
Forgot to talk about this in the meeting today, but I'll implement some kind of prod warning before merging, @XanderVertegaal. I don't think absolute prevention is reasonably possible. Thanks for the input!

PR #169, opened 2026-03-05
Feat/tiny mce step info text

### LINKED ISSUE #158: StepInfo object should have rich text editor
We'll need a rich text editor to write up stepinfo. Then we can do links, bold and italics easily, which we'll need. We can then also get rid of the Step info questions object.

We can use TinyMCE widget for this which is part of the Django Shared Core.

EdoStorm96:
fixes #158 

So now we have a richtext editor for StepInfo. Very nice for when we make the form. Though I do still have some doubts about security ...

I did disable the part of the TinyMCE widget where you can enter raw html, but I am still asking vue to display raw html. I think, as for now this widget is in the admin pages only, it should be fine, but we should look into how hackable this widget is if we ever want to give it to users.

I also removed StepInfoQuestions and made StepInfoText a 1-to-1 with a Step. 

I suggest giving this a go in the admin interface. It is quite nice to use.


backend/form/types/StepType.py
```diff
@@ -1,11 +1,10 @@
-from graphene import ID, Int, List, NonNull, ObjectType, ResolveInfo, String
+from graphene import ID, Field, Int, List, NonNull, ObjectType, ResolveInfo, String
 
 from django.db.models import QuerySet
```
XanderVertegaal: Unused import.

backend/form/admin.py
```diff
@@ -40,16 +42,33 @@ class SubstepInline(admin.StackedInline):
...
-
-
 class StepInfoTextInline(admin.StackedInline):
     model = StepInfoText
     extra = 0
-    fields = ("text",)
+    fields = ("text_en", "text_nl")
+
+    class Media:
+        js = (
```
XanderVertegaal, line 34: Would it be worthwhile to outfactor the Media class and the TinyMCEWidget and to use them in both `StepInfoTextInline` and `StepInfoTextAdmin`?

backend/main/management/commands/create_dev_data.py
```diff
@@ -175,24 +173,12 @@ def generate_form_info_text() -> None:
...
-        def generate_form_info_questions() -> None:
-            for _ in range(self.faker.random_int(1, 3)):
-                StepInfoQuestion.objects.create(
-                    step=step,
-                    text_nl=self.faker_nl.sentence(),
-                    text_en=self.faker_en.sentence(),
-                    link=self.faker.url(),
-                )
-
         choice = self.faker.random_element(["text", "questions", "both"])
```
XanderVertegaal: You can replace this with `self.faker.pybool()` or something like that and adapt the conditional below.

backend/myresearch/settings/generic_settings.py
```diff
@@ -44,6 +44,8 @@
         "graphene_django",
         "modeltranslation",
         "django_extensions",
+        # CDH-Django-core libraries
```
XanderVertegaal: The comment here does not really add a lot.

backend/form/models.py
```diff
@@ -87,16 +87,8 @@ def __str__(self):
...
 
-class StepInfoQuestion(models.Model):
-    step = models.ForeignKey(
-        Step, on_delete=models.CASCADE, related_name="info_questions"
-    )
-    text = models.CharField(max_length=200)
-    link = models.URLField(max_length=200)
-
-
 class StepInfoText(models.Model):
```
XanderVertegaal, line 90: You could even go one step further: since there is no real distinction between `StepInfoText` and `StepInfoQuestion`, maybe this model can simply be called `StepInfo`? I'll leave this up to you; this is not required.

backend/form/types/StepType.py
```diff
@@ -27,17 +26,10 @@ class StepType(ObjectType):
         lambda: NonNull(StepType),
         required=True,
     )
-    info_questions = List(NonNull(StepInfoQuestionType), required=True)
-    info_texts = List(NonNull(StepInfoTextType), required=True)
+    info_text = Field(StepInfoTextType, required=True)
```
XanderVertegaal: This is incorrect: not every `Step` will have a `StepInfoText` (although the opposite is true).


```suggestion
    info_text = Field(StepInfoTextType)
```

XanderVertegaal: Also the resolver below will have to be adjusted:

```py
    @staticmethod
    def resolve_info_text(parent, info: ResolveInfo) -> StepInfoText | None:
        if not parent.step_id:
            return None # Note that .objects.none() returns a queryset.
        try:
            return StepInfoText.objects.get(step_id=parent.step_id)
        except StepInfoText.DoesNotExist:
            return None
```

frontend/components/shared/FormSideBar.vue
```diff
@@ -22,38 +16,14 @@ const props = defineProps<{
...
-                    <li
-                        v-for="(info, index) in stepInfo.infoTexts"
-                        :key="index"
-                    >
-                        {{ useTranslateableAttribute(info, "text") }}
-                    </li>
-                </ul>
+            <div v-if="stepInfo">
+            <strong>{{ $t("Additional Information") }}</strong>
+                <div v-html="useTranslateableAttribute(stepInfo.infoText, 'text')"></div>
```
XanderVertegaal, line 26: This will have to be adapted too if `stepInfo.infoText` is made optional.

```vue
            <div v-if="stepInfo.infoText">
```

XanderVertegaal (commented):
It looks and feels very nice! I don't think there are any serious security issues here as long as we are ones who take care of the input. (I don't think we plan to let other users into Django Admin.)

You will have to remove `required=True` on `Step.info_text` and change the resolver + frontend accordingly (see the comments) but after that it's good to merge as far as I'm concerned.

backend/form/models.py
EdoStorm96: Mmm ... I do agree this would be better ... But I'm going to leave it.

backend/main/management/commands/create_dev_data.py
```diff
@@ -173,11 +173,8 @@ def generate_form_info_text() -> None:
                     text_en=self.faker_en.paragraph(),
                 )
 
-        choice = self.faker.random_element(["text", "questions", "both"])
-
-        if choice == "text":
-            generate_form_info_text()
-        else:
+        breakpoint()
```
XanderVertegaal: Is this breakpoint a leftover from a previous debugging session?

XanderVertegaal (approved):
LGTM!

PR #172, opened 2026-03-10
Feature/vwr questions

XanderVertegaal:
Provides fixtures for the VWR form. As discussed, I tried removing the PK on the top-level form so you can easily make new copies, but Steps refer to a form object, so removing the PK would break the foreign keys.

Bonus features:
- QuestionCondition.trigger_value (a JSONField) now has `blank=True`, so `{}` is recognised as a valid JSON value.
- You can now set the Dutch label too for questions created as part of the StepForm in QuestionInline. (Previously, only the English label could be set.)
- For SelectQuestions, we now simply use a JSON with 'value' instead of 'option_ids' as the key, and a comma-separated string of IDs, e.g. `{"value": "12,13"}`.
- In Django Admin, StepForm now allows you to determine the order of substeps (similar to questions).
- For repeated questions/steps, the minimum number of repetitions was 1; now it is 0.

Still to do:
- The fixtures do not seem to respect the `_order` field when the data is loaded in. Some questions are not ordered correctly. This can/should easily be corrected on the server when we deploy.
- You should be able to determine the order of options in SelectQuestions. This is currently not supported yet.
- The prototype has several questions/steps whose descriptions contain lists or other advanced formatting. We may want to implement TinyMCE for these description fields too.
- The "default selected" option for SelectQuestions does not seem to be respected.
- Caching issue with duplicated steps: some answers seem to bleed over from one repeated step to the next. This may be a frontend-only bug: there are Apollo warnings in the frontend.
- Questions that were filled in and subsequently hidden retain their value. As a result the form may have "hidden" values that render it invalid.


EdoStorm96 (approved):
Looks great! Approved.

I found that we can fix the issue with the question order by just moving things around in the json file to be in the correct order. Have a look at #180.

PR #180, opened 2026-03-12
fix: change order of fixture

### LINKED ISSUE #178: Order of questions is not preserved on loading fixtures
The questions in the fixtures of #172 have an `_order` field set, but this order does not seem to be respected when you run `python manage.py loaddata`. You have to set the order of questions manually in `StepAdmin`.

EdoStorm96:
fixes #178

I found that by just moving the json around, the order already gets fixed. I've gone through all the questions to see if they correspond with the order and they're all fixed now.

I'd say give it a test, but it is working for me :)

### DIFF
--- a/backend/form/fixtures/vwr.json
+++ b/backend/form/fixtures/vwr.json
@@ -370,21 +370,6 @@
     },
     {
         "model": "form.basequestion",
-        "pk": 11,
-        "fields": {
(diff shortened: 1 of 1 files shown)


XanderVertegaal (approved):
Wow, this works! Interesting to see that the order of the fixtures matters so much. Thanks for figuring this out!

I will merge both of our PRs right away :)

PR #189, opened 2026-03-18
Feature/reference numbers

### LINKED ISSUE #126: Reference number for Studies
We've decided to not include chamber or faculty info in the reference number, because good reference numbers are constant and we may not have that information right from the start.

We want to include two digits for a year. And four incrementing digits for each study, so we encode some amount of information in the reference number for all to see.

Example: MR-26-0001

We areen't yet taking revisions into account here. We may still do that in the future. 

For example: MR-26-0001-01

EdoStorm96:
fixes #126 

I've done it as described in the issue. I think for now this is sufficient. The ref num gets generated upon creation of the study. I've created a counter model for each year so we can keep track of where we are and avoid it breaking under concurrency.

Works fine and I've implemented it in the frontend where it is relevant.


frontend/pages/studies/[study_id].vue
```diff
@@ -91,7 +92,7 @@ const studyStatus = computed(() =>
                         <!-- Main Content -->
                         <div class="col me-5">
                             <h1>
-                                2025-{{ randomNumber100to1000() }} -
+                                {{ study?.reference }} -
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id].vue]: This is going to show `undefined - undefined` if there is no study. I don't think it should happen, but just in case I would use a fallback here.

The `randomNumber100to1000` can also be cleaned up if you're not using it elsewhere.

frontend/pages/studies/[study_id].vue
```diff
@@ -19,6 +19,7 @@ const GET_STUDY = graphql(`
             }
             id
             title
+            reference
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id].vue]: Nitpick: it is customary to query simple fields before related ones first, so:

```ts
const GET_STUDY = graphql(`
    query GetStudy($id: ID!) {
        study(id: $id, mrPermission: "View") {
            id
            title
            reference
            createdBy {
                fullName
                id
                email
            }
```

(This isn't a change you made in this PR; just something I'm noticing now.)

backend/research/models.py
```diff
@@ -37,3 +41,20 @@ def status(self):
         return self.status_changes.last().status
 
     objects = StudyManager()
+
+    def save(self, *args, **kwargs):
+
```
XanderVertegaal: (Remove the whiteline)

```suggestion
```

backend/research/models.py
```diff
@@ -37,3 +41,20 @@ def status(self):
...
+        if not self.reference:
+            with transaction.atomic():
+                year = timezone.now().year % 100  # 2026 -> 26
+
+                counter_obj, _ = YearCounter.objects.select_for_update().get_or_create(
```
XanderVertegaal, line 82: Oh wow, today I learned about `select_for_update`! I'm not sure we'll ever have a race condition like this, but it's good to have this in place nonetheless.

backend/research/other_models/utils.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+
+class YearCounter(models.Model):
+    """
+    A helper model for generating reference numbers for studies. Keeps a counter
+    for each year, that get incremented when a new Study is created.
```
XanderVertegaal:
```suggestion
    for each year that gets incremented when a new Study is created.
```

XanderVertegaal (approved):

Issue #187, opened 2026-03-18
remove missing models check from dev data


EdoStorm96:
It was a cool idea to check for missing models in create_dev_data, but it's not really practical. I think this should probably be removed.

PR #201, opened 2026-03-22
Outfactor dev data generation completeness check to test

XanderVertegaal:
The check whether all models are accounted for in `create_dev_data` no longer happens in the command itself, but in an individual test. The same test also checks whether `create_dev_data` completes successfully, so we are notified if one of our PRs threatens to break it.

Instead of adding a model to a list if you _have_ implemented a dev data generator for it, you now have to opt out by adding models for which it is _not_ represented in `create_dev_data`. On all models not present in the list, `objects.exists()` is called to check whether objects actually exist (instead of taking the developer's word for it).

### DIFF
--- a/backend/form/services/form_evaluator_test.py
+++ b/backend/form/services/form_evaluator_test.py
@@ -648,8 +648,8 @@ class TestFormEvaluatorSubmission:
         newer = UserFormSubmission.objects.create(user=test_user, form=form)
 
         evaluator = FormEvaluator(form, test_user)
-
-        assert evaluator.submission == newer
+        assert evaluator.submission
+        assert evaluator.submission.pk == newer.pk
 
     def test_get_user_answer(self, form, test_user, trigger_question):
         """Test get_user_answer retrieves correct answer."""
--- a/backend/main/management/commands/create_dev_data.py
+++ b/backend/main/management/commands/create_dev_data.py
@@ -1,7 +1,6 @@
 from tqdm import tqdm
 from faker import Faker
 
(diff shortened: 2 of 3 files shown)


backend/form/services/form_evaluator_test.py
```diff
@@ -648,8 +648,8 @@ def test_returns_latest_submission(self, form, test_user):
         newer = UserFormSubmission.objects.create(user=test_user, form=form)
 
         evaluator = FormEvaluator(form, test_user)
-
-        assert evaluator.submission == newer
+        assert evaluator.submission
```
XanderVertegaal, line 651: This test was flaky for me. I think it failed when doing strict comparison of the `created_at` field (and there was some millisecond discrepancy or something). This is safer.

backend/main/management/commands/create_dev_data.py
```diff
@@ -136,6 +113,7 @@ def _generate_substeps(step: Step, depth: int) -> None:
             for _ in tqdm(
                 range(num_substeps),
                 desc=f"Generating substeps for step {step.name[:10]} (depth: {depth})",
+                disable=options["silent"],
```
XanderVertegaal, line 116: `disable` on `tqdm` determines whether output is printed to stdout.

EdoStorm96 (approved):
Yep, great idea! I was already thinking about removing this from create_dev_data, but doing it like this is very nice!

fixes #187

miggol (approved):
Prefect!

PR #212, opened 2026-03-31
Rich text formatting in question/step descriptions

### LINKED ISSUE #175: Rich text formatting in question/step descriptions
In the VWR prototype, some descriptions have numbered lists. I can imagine others using boldface for emphasis and other typesetting features too.

With TinyMCE implemented for StepInfo in #169, we can extend this feature to the description fields as well.

RoordinkFrank:
fixes #175

Do note that the original description fields still have the .css muted option enabled.  That happens on top any formatting done in admin.
Do also note that links in mcedit are relative by default unless you add www/html etc. Personally I think that's really helpful.

--------
known warning. I have tried to suppress this but es lint suppression warnings don´t work in HTML without an additional plugin.
This warning is also something Leon and Desiree need to be aware of if we want to give them access to this widget.
image


frontend/components/shared/MRForm.vue
```diff
@@ -61,7 +61,7 @@ function hasErrors(question: QuestionWithValue): boolean {
 
 <template>
     <h2>{{ useTranslateableAttribute(step, "name") }}</h2>
-    <p>{{ useTranslateableAttribute(step, "description") }}</p>
+    <div v-html="useTranslateableAttribute(step, 'description')" />
```
EdoStorm96: Here, you replace a `<p>` with a `<div>`. I think this is probably the ideal way to do it, but I think you should add a proper closing tag.
```suggestion
    <div v-html="useTranslateableAttribute(step, 'description')" ></div>
```
However, do notice, that when you generate a fresh form, this text looks a bit weird. This is because in create_dev_data we just generate plain strings, with no html, so this does not have a `<p>` tag at all.

In production we always will expect our data to be passed through our tinymce editor, and therefore have some basic html formatting, so I think a div is the cleanest tag to use, to avoid having double `<p>` tags.

EdoStorm96 (changes requested):
Good job! This is pretty much what I wanted. I still have some things I am noticing, that I'd like you to change though.

- You removed the closing tags from all html element. You made it just `<p/>` instead of `<p></p>`. Apparantly vue (or something) does turn this into proper html, but I'd like you to bring back closing tags.
- You've kept all the description `<p>` elements (except for the Step.description. See my comment on MRForm). While this does ensure that plain strings will be formatted nicely, once you edit it with tinymce, we end up with p tags wrapped in p tags. I think this is a bit ugly, so I'd like you to change all the description to use the div element.
- However, when you do this, we end up with ugly unformatted text, coming from create_dev_data. I noticed this on the step description, because you already use a div there.

So what I'd like you to do, is alter our create_dev_data script, so that all descriptions are at least wrapped in `<p>` tags. Something like this:

create_dev_data.py, line 185-186
```
                "description_nl": f"<p>{self.faker_nl.paragraph()}</p>",
                "description_en": f"<p>{self.faker_en.paragraph()}</p>",
```
Also do this for step descriptions. Then our mock data will reflect our eventual data better, and things will be looking good.

We don't mock step info yet, so there we'll also have to do this, but that is something for another issue/PR.

frontend/components/shared/MRForm.vue
XanderVertegaal: Fun fact: omitting closing tags is valid HTML for many elements (which is probably why our editors don't complain about it). [More discussion here](https://stackoverflow.com/questions/37030607/html-style-guide-google-vs-w3schools-omit-optional-tag?noredirect=1&lq=1) Having said this, I do think we should always close our tags (unless they're self-closing, like `<br />` or void, like ``).

RoordinkFrank:
Be aware, the create_dev_data only change the <p> in well dev data. The my research fixture do not have this until the next time a question is changed.

RoordinkFrank:
Merge went wrong or I have been missing something.

image

Edit: fix found, it was a code from a recent branch which did not yet have the v-html, which is logical as it only gets added in this branch.

EdoStorm96 (commented):
> Be aware, the create_dev_data only change the
> 
> in well dev data. The my research fixture do not have this until the next time a question is changed.

This is a very good point. As we discussed, it would be really nice, if you can add basic html formatting to out fixture as well. Then we are sure everything will get displayed correctly.

RoordinkFrank:
VWR created with` sudo docker exec -it mr-django-dev sh` ` python manage.py dumpdata --indent 4 form --exclude form.questionresponse --exclude form.userformsubmission > form/fixtures/vwr.json`, Does not work in one go for some reason.

Once again if someone has time time please reformat the code so that the VWR.json is way easier to validate. This usually can be done by right clicking on a file in an editor and reformat. My editor (with current options) thinks this is the right reformat. Difference is an extra tab.

frontend/components/form/MRForm.vue
```diff
@@ -45,6 +45,8 @@ useConditionalQuestionsWatcher(questionsWithConditions, () => {
 </script>
 
 <template>
+    <h2>{{ useTranslateableAttribute(step, "name") }}</h2>
```
EdoStorm96: The title and description of a step are shown twice now. I think this might have happened when merging with develop. Maybe ask @XanderVertegaal if the place where step description and name are shown have been changed recently? They are also shown in FormWrapper in this branch.

backend/form/fixtures/vwr.json
```diff
@@ -266,7 +266,7 @@
             "description_en": "",
             "description_nl": "",
             "required": false,
-            "_order": 1
+            "_order": 4
```
EdoStorm96, line 458: So more then just the description has been changed here ... Would you mind if I had a try at updating our fixture?

EdoStorm96 (commented):
Mmm ... Some weird stuff has happened with the branch. The fixture features some unexpected changes and Step title and description are shown twice now. Would you mind if I had a try at updating the fixture?

frontend/components/form/MRForm.vue
XanderVertegaal: I think I've lifted them to the FormWrapper component:

image

backend/form/fixtures/vwr.json
```diff
@@ -517,9 +517,9 @@
...
             "text_nl": "Documenten uploaden",
             "step": 12,
-            "description": "- DPIA (Data Protection Impact Assessment\r\n- Privacy scan, pre-GBEB, DPIA Light, Privacy review, or other abbreviated risk assessment\r\n- Other relevant privacy-related documents, data management plan (DMP)\r\n- Privacy statement",
-            "description_en": "- DPIA (Data Protection Impact Assessment\r\n- Privacy scan, pre-GBEB, DPIA Light, Privacy review, or other abbreviated risk assessment\r\n- Other relevant privacy-related documents, data management plan (DMP)\r\n- Privacy statement",
-            "description_nl": "-     DPIA (Data Protection Impact Assessment, ook wel GBEB genoemd)\r\n-     Privacy scan, pre-GBEB, DPIA Light, Privacy review of andere verkorte risico-inventarisatie\r\n-     Andere relevante, privacy-gerelateerde documenten, datamanagementplan (DMP)\r\n-     Privacyverklaring",
+            "description": "<p>- DPIA (Data Protection Impact Assessment - Privacy scan, pre-GBEB, DPIA Light, Privacy review, or other abbreviated risk assessment - Other relevant privacy-related documents, data management plan (DMP) - Privacy statement</p>",
```
XanderVertegaal: This should probably be formatted as an unordered list.

backend/form/fixtures/vwr.json
```diff
@@ -1785,7 +1785,7 @@
     },
     {
         "model": "form.truefalsequestion",
-        "pk": 12,
+        "pk": 21,
```
XanderVertegaal: Why are these pks different?

backend/form/fixtures/vwr.json
```diff
@@ -2229,9 +2229,9 @@
...
-            "description_nl": "In deze stap geef je op welke categorieën betrokken mensen een rol spelen in je onderzoeksproject. \r\n\r\nPer categorie betrokken mensen vragen we twee dingen:\r\n1. Waarom het noodzakelijk is om persoonsgegevens te verzamelen.\r\n2. Welke specifieke categorieën persoonsgegevens je gaat verwerken.\r\n\r\nAls je in de vorige vraag meerdere categorieën betrokkenen hebt opgegeven, wordt je dus gevraagd om voor elk van die categorieën informatie op te geven.",
+            "description": "<p>In this step you will be asked to indicate which categories of data subjects play a role in your research project. For each category of data subjects, we ask two questions: 1. Why is it necessary to collect personal data? 2. Which specific categories of personal data you will be processing? If you submitted multiple categories of data subjects in the previous question, you will be asked to provide information for each of those categories.</p>",
```
XanderVertegaal: An ordered list would be good here.

XanderVertegaal (commented):
In addition to @EdoStorm96's comments (about the seemingly changed question order and the duplicated title + description), you might want to consider formatting a few descriptions in the VWR fixtures as lists.

Otherwise, the code changes look reasonable to me! 👍

frontend/components/form/MRForm.vue
RoordinkFrank: This seems to be the case. Should be fixed now.

backend/form/fixtures/vwr.json
RoordinkFrank: I did regenerate the fixtures from scratch. The pk's did change, why no idea.
`python manage.py dumpdata --indent 4 form --exclude form.questionresponse --exclude form.userformsubmission > form/fixtures/vwr.json`

RoordinkFrank: new and old pkś and order look exactly the same with for example: (vrw.json line 160)
Image

RoordinkFrank:
fixtures have been updated with the new command. I still see some of the fixtures switching around but other than that their references stay the same.

PR #213, opened 2026-03-31
Feature/prod fixes

miggol:
Collection of random fixes required to get prod running on servers


compose.yaml
```diff
@@ -149,7 +162,7 @@ services:
     extra_hosts:
       - "host.docker.internal:host-gateway"
     environment:
-      - DJANGO_SECRET_KEY=django-insecure-lb=q@u4df-x0th(5u%$eye_ti#etst+5z+%2=lrh$$le3&v_y$
+      - DJANGO_SECRET_KEY=django-insecure-wa2AengahShofo8tieR6
```
miggol, line 163: Sorry about this, might have to rebuild your dev-idp. But this is YAML and we can't have special characters in our values without escaping them...

backend/myresearch/settings/dev.py
```diff
@@ -1,4 +1,7 @@
-from .utils import discover
+import logging
+from .utils import discover, discover_list
+
+logger = logging.getLogger(__name__)
```
miggol, line 4: Please use this syntax wherever you need a logger from now on :)

EdoStorm96 (approved):
I think it looks fine! If you can fix the problem with the check, I think this is good to merge.

PR #214, opened 2026-03-31
Feature/generate dev question responses

### LINKED ISSUE #194: Mock UserFormSubmission data
Update `create_dev_dataset` with realistic UserFormSubmission data. This is going to be tricky, since it depends on a dynamic MRForm, which is dynamic, but it should be doable.

XanderVertegaal:
Closes #194 

As discussed a few weeks ago, this adjusts the `create_dev_dataset.py` script. If `create_dev_dataset` is run without any flags, it will generate a random form with random steps and random questions. If you run `create_dev_dataset --vwr`, on the other hand, this will load the existing processing registry fixtures instead.

Regardless, the script will then take the generated form (random or procreg) and create 1-4 random studies and 1 submission for each study  for each user in the database. This takes a while (2:20 mins when I ran it), which is why I opted for not generating more than 1 submission per study. We can perhaps make this a bit more performant in the future, but for now this will do.


EdoStorm96 (commented):
Besides merge conflicts, this all seems to be working well. Nice!

This did lead me to discover (as I've discussed with you) that a fully loaded USF cannot (with my patience) be loaded in the admin interface due to the heaviness of loading all the QuestionResponseInline's for one USF ... If you would fix that in this PR, it would be nice, but it is not really related to this, so it is not a must for me.

I would appreciate if you could bring this up to date with develop, so I can see the generated responses in action. Would you notify me when you've done this so I can have a quick last look?

XanderVertegaal:
> I would appreciate if you could bring this up to date with develop, so I can see the generated responses in action. Would you notify me when you've done this so I can have a quick last look?

Done! I have added a bit of prefetching to make the query fetching the UFS more performant. Now the Django Admin page loads within seconds. Have a look!

EdoStorm96 (approved):
Wow, crazy how much time this saves! Nice one.

backend/main/management/commands/create_dev_data.py
```diff
@@ -263,29 +296,115 @@ def _get_all_substeps(step: Step) -> list[Step]:
...
+        return {"value": [option.pk for option in selected_options]}
+
+    def _generate_true_false_answer(self, question: TrueFalseQuestion) -> dict:
+        return {"value": self.faker.pybool()}
+
+    def _generate_date_answer(self, question: DateQuestion) -> dict:
+        start_date = "today" if question.future_only else "-5y"
+        end_date = "+5y" if question.future_only else "today"
+        return {
+            "date": self.faker.date_between(
```
EdoStorm96:
```suggestion
            "value": self.faker.date_between(
```
This was causing all the answers to be scrambled. I'm not sure how this caused it ... But this does fix the problem :)

PR #215, opened 2026-04-02
Feat/UI study creation

EdoStorm96:
So here is a PR which enables the creation of studies (and related models) through our UI. But is also adresses several other related problems:

Sidenote, I've made use of copilot for this PR (hence my productivity ;p), so have a careful look please.

- authentication checking and redirecting to login:
  - I did not want users to be able to create a study without being authenticated, as we need a user to create a study (and related models)
  - most important file related to this is auth.client.ts. This is a global nuxt plugin, which runs on every page and checks if a user is authenticated and redirects to login if not. You can bypass this by adding `public: true` to `definePageMeta` of a page, like I've done on the homepage.
  - I'm still not super comfortable with nuxt/vue/ts stuff, so @XanderVertegaal please examine this critically.
- I've separated create and update responsibilities.
  - As discussed, I've had a change of heart about how studies should be created, and I actually prefer creating everything beforehand. This happens in `create_study.py`, which gets called by `CreateStudyMutation`.
  - We now create a Study, a submission and an inital draft `StatusChange` all at the same time.
  - this mutation gets called from localhost:5000/procreg/start/. If you call the mutation, you get redirected to the study's detail page.
  - I've written tests for create_study and have rewritten the update_submission tests. (thx Claude)
- I've had to reintroduce `Review` models, as these were accidentally removed in a previous migration (`backend/research/migrations/0003_remove_reviewround_reviewed_form_and_more.py`). 
  - We'll not be using the review models for the MVP, but I did want the `StatusChange` model back, so I brought them all back.
  - I did leave the remove migration, as it was already in develop with other migrations depending on it.
- There are now admin pages for Studies and StatusChanges (thx Claude)
- I've reorganized the navbar, to reflect our new functionality.

So it's quite a lot, but there is some nice functionality here. Maybe we should discuss in a meeting how we feel about using copilot ...


EdoStorm96:
Alright, after addressing all comments made by copilot, I realized, I did not fully implement the changes to update_submission, especially the effects this has on how we handle FormQueries (like in FormEvaluator we still have the ability to create a submission). But I will leave this for a later moment. The part of the PR that is important to review right now (the creation of a study) is finished and ready to be reviewed. I will fork from here to fix the actual form updating and querying part but that is currently broken and inaccessible ...

EdoStorm96:
Okay, I've adressed the things I mentioned in my latest comment. What I've added:
- The FormEvaluator just receives a submission, as that should always already exist now
- I've updated our tests to reflect this.
- I've changed the path for editing a form, so it is now `procreg/[submission_id]/[slug]`. I've updated the form query to also work like this.
- the continue button now work on the study page

Sorry for the giant PR. Things kept snowballing ...

backend/form/mutations/UpdateUserFormSubmission.py
```diff
@@ -25,31 +25,19 @@ def mutate(
     ):
         user: User = info.context.user
         responses = getattr(user_form_input, "responses", [])
-        create_study_flag = getattr(user_form_input, "create_study", False)
 
         try:
-            submission = update_or_create_submission(
+            submission = update_submission(
```
XanderVertegaal: Unused variable (same is true for `responses` on line 27).

backend/form/services/update_submission.py
```diff
@@ -3,21 +3,14 @@
 from form.mutations.utils.inputs import UserFormInput
 
 
-def update_or_create_submission(
-    user: User, user_form_input: UserFormInput
-) -> UserFormSubmission:
-    # Usually we can use user_form_input.form_config_id or getattr(user_form_input, "form_config_id").
+def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSubmission:
```
XanderVertegaal, line 6: `user` is not used.

backend/form/services/update_submission.py
```diff
@@ -3,21 +3,14 @@
...
-    form_config_id = user_form_input.get("form_config_id", None)  # type: ignore
 
-    if submission_id:
-        current_submission = UserFormSubmission.objects.get(id=submission_id)
-    else:
-        current_submission = UserFormSubmission.objects.create(
-            user=user, form_id=form_config_id
-        )
+    current_submission = UserFormSubmission.objects.get(
+        id=user_form_input.get("submission_id")
```
XanderVertegaal: Remember that anyone can send GraphQL requests to our endpoint and write their own mutations with their own variables, especially if we have a public page and our GraphQL endpoint is currently unprotected by CSRF (you can see that `csrf_exempt` is used in `urls.py`). In the future I'd like to suggest that our GraphQL endpoint requires authentication. (The DIAPP has this in place, for example.) 

For now it would already help if you could check whether the user making the request actually owns the submission they're trying to update. Another good addition would be to wrap this in a `try...except...` block.

backend/form/mutations/utils/inputs.py
```diff
@@ -9,7 +9,5 @@ class ResponseInput(InputObjectType):
 
 
 class UserFormInput(InputObjectType):
-    submission_id = ID()
-    form_config_id = ID(required=True)
+    submission_id = ID(required=True)
     responses = List(NonNull(ResponseInput), required=True)
-    create_study = Boolean(required=False, default_value=False)
```
XanderVertegaal, line 15: `Boolean` is now an unused import.

backend/form/queries.py
```diff
@@ -1,33 +1,33 @@
...
     form = Field(
         UserFormType,
+        submission_id=ID(required=True),
         description="Retrieves the user's submission for a specific form.",
     )
 
     @staticmethod
-    def resolve_form(root, info: ResolveInfo) -> Optional[UserFormType]:
+    def resolve_form(
+        root, info: ResolveInfo, submission_id: int
```
XanderVertegaal: I think Graphene `ID`s are technically mapped to `str`, not `int`, but I could be wrong and it should not make a difference.

backend/form/queries.py
```diff
@@ -1,33 +1,33 @@
...
+    ) -> Optional[UserFormType]:
         user: User = info.context.user
         if not user.is_authenticated:
             return None
 
-        # Get the latest form config/template (for dev purposes only)
-        form_config = MRForm.objects.order_by("-created_at").first()
+        # TODO: implement permissions
+        submission = UserFormSubmission.objects.get(id=submission_id)
```
XanderVertegaal: Maybe a quick check to see whether the user is the one that owns the submission? I know we're still actively developing and the application contains no real data, but it does feel uneasy to see that any user can request _any_ submission.

backend/research/admin.py
```diff
@@ -1,3 +1,62 @@
...
+    def current_status(self, obj):
+        return obj.status
+
+    def has_add_permission(self, request):
+        """Disable adding studies through admin"""
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        """Disable deleting studies through admin"""
+        return False
```
XanderVertegaal, line 53: Why? I don't see any obvious usecase for it, but is there something wrong with deleting/adding studies via Django Admin?

frontend/components/shared/FormWrapper.vue
```diff
@@ -12,142 +12,136 @@ import type { UpdateUserFormSubmission } from "~/generated/gql/graphql";
...
 
 const queried = computed(() => props.queriedForm);
 const { formObject, validationRules } = useFormState(queried);
 
 const UPDATE_USER_FORM = graphql(`
-    mutation SaveFormSubmission(
-        $submissionId: ID
-        $formConfigId: ID!
-        $responses: [ResponseInput!]!
+  mutation SaveFormSubmission(
```
XanderVertegaal: The indentation is a bit strange here. Could you run prettier again?

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,74 +1,105 @@
...
+);
+
+const { result } = useQuery<GetFirstSlugQuery>(GET_FIRST_SLUG, queryVariables);
+
+const slug = computed<string | null>(() => {
+  const firstStep = result.value?.form?.steps[0];
+  return firstStep?.slug || null;
+});
+
+const continue_url = computed(() =>
```
XanderVertegaal:
```suggestion
const continueUrl = computed(() =>
```

frontend/pages/studies/[study_id].vue
```diff
@@ -10,108 +10,112 @@ import StudyProgessBar from "../../components/study_detail/StudyProgessBar.vue";
...
+                >.
+              </p>
+              <AvailableActions
+                :study-status="studyStatus"
+                :submission-id="study?.latestSubmissionId"
+              />
+            </div>
+            <!-- Progess bar -->
+            <div class="col-2">
+              <StudyProgessBar :study-status="studyStatus" />
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id].vue]:
```suggestion
                        <div v-if="study" class="col me-5">
```

If you do this, you can avoid seeing `undefined` in your template (e.g. for `study?.title`) if there is no study. You can then also assume in `AvailableActions` that `submission-id` and `study-status` are never undefined, which makes that component a bit neater.

frontend/pages/studies/[study_id].vue
```diff
@@ -10,108 +10,112 @@ import StudyProgessBar from "../../components/study_detail/StudyProgessBar.vue";
...
 function randomNumber100to1000(): number {
-    return Math.floor(Math.random() * (1000 - 100 + 1)) + 100;
+  return Math.floor(Math.random() * (1000 - 100 + 1)) + 100;
 }
 
 // If study is even, it is a draft. If it is odd, it is in the review phase
 
 const studyStatus = computed(() =>
-    Number(study.value?.id) % 2 === 0 ? "draft" : "review",
+  Number(study.value?.id) % 2 === 0 ? "draft" : "review",
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id].vue]: If there is no study, `Number(study.value?.id) % 2` will be `NaN`, so you'd see `review`. Not fatal, but not correct either.

frontend/pages/studies/[study_id].vue
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id].vue]: (I guess this won't be here after merging the reference number PR, but good to know: props are always kebab case in Vue.)

```suggestion
                :random-date-past-year="randomDatePastYear()"
```

Comment (XanderVertegaal) [reply frontend/pages/studies/[study_id].vue]: If you do this, you can also use `v-else` to add a simple "No study found" message.

frontend/pages/studies/[study_id].vue
```diff
@@ -10,108 +10,112 @@ import StudyProgessBar from "../../components/study_detail/StudyProgessBar.vue";
 // retrieve study
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id].vue]: This comment is a bit superfluous IMO.

frontend/pages/studies/[study_id].vue
```diff
@@ -10,108 +10,112 @@ import StudyProgessBar from "../../components/study_detail/StudyProgessBar.vue";
...
 
 const route = useRoute();
 
 const {
-    result: studyResult,
-    loading,
-    error,
+  result: studyResult,
+  loading,
+  error,
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id].vue]: Error is not used.

frontend/plugins/auth.client.ts
```diff
@@ -0,0 +1,40 @@
+export default defineNuxtPlugin(() => {
+  addRouteMiddleware(
+    "global-auth",
+    async (to, from) => {
+      // Check if page is marked as public
+      if (to.meta.public) {
+        return;
+      }
```
XanderVertegaal: Please apply Prettier here as well for consistent tabs/spaces.

XanderVertegaal: Nice, quick and easy 👍

frontend/plugins/auth.client.ts
```diff
@@ -0,0 +1,40 @@
+export default defineNuxtPlugin(() => {
```
XanderVertegaal: I think it is more idiomatic to put this middleware in `/frontend/middleware/01.auth.global.ts` The `middleware` directory is a Nuxt feature: all modules there are registered as middleware so you don't have to define a plugin for them, and the `.global` indicates that it should run on every page load (not just one or two routes). Files in `/middleware` are run in alphebetical order, so it is customary to prefix the names with `01`, `02` etc. to ensure proper ordering. I guess you don't need the `global: true` anymore if you use this structure. You can then replace `addRouteMiddleware` (which seems to be more appropriate if you want to add middleware on the fly, as part of a component, for instance) with the following:

```ts
export default defineNuxtRouteMiddleware(async (to, from) => {
    ...
```

frontend/pages/procreg/start.vue
```diff
@@ -0,0 +1,81 @@
...
+      return;
+    }
+
+    if (
+      result.data.createStudy?.errors &&
+      result.data.createStudy.errors.length > 0
+    ) {
+      useNotification("Failed to create study", "danger");
+      return;
+    }
```
XanderVertegaal:
```suggestion
        if (result.data.createStudy?.errors?.length) {
            useNotification("Failed to create study", "danger");
            return;
        }
```

frontend/pages/procreg/start.vue
```diff
@@ -0,0 +1,81 @@
...
+        @click="handleCreateStudy"
+        :disabled="loading"
+        class="btn btn-primary btn-lg"
+      >
+        <span
+          v-if="loading"
+          class="spinner-border spinner-border-sm me-2"
+        ></span>
+        {{ loading ? $t("Creating...") : $t("Register new study") }} >>
+      </button>
```
XanderVertegaal: Event handlers are usually the last attributes on an element.

frontend/pages/procreg/start.vue
```diff
@@ -0,0 +1,81 @@
...
+      result.data.createStudy.errors.length > 0
+    ) {
+      useNotification("Failed to create study", "danger");
+      return;
+    }
+
+    if (result.data.createStudy?.study?.id) {
+      useNotification("Study created successfully", "success");
+      // Redirect to the study detail page
+      router.push(`/studies/${result.data.createStudy.study.id}`);
```
XanderVertegaal: I'm getting an ESLint error here: `Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.`


```suggestion
            void router.push(`/studies/${result.data.createStudy.study.id}`);
```

frontend/pages/procreg/start.vue
XanderVertegaal:
```suggestion
        if (result.data.createStudy?.errors?.length) {
```

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,74 +1,105 @@
...
+  slug.value ? `/procreg/${props.submissionId}/${slug.value}` : "",
```
XanderVertegaal: Please handle the case where `submissionId` is undefined (or make sure that it is never undefined, cf. my comment about this elsewhere).

XanderVertegaal (commented):
Works like a charm, very well done! I have a few suggestions and comments, as always; the only one I'd really like you to consider is to add a user check to the submission update mutation so as to avoid that _anyone_ can update any submission.

Also, it would be great if you could add a linting pass to this PR (`prettier . --write`). A lot of files have meaningful changes buried underneath a general change to the number of spaces used for indentation.

backend/research/admin.py
EdoStorm96: Mmm ... For me this feels like something that we should only allow with the correct permissions in place, which we can enforce in the frontend. Furthermore, when it comes to adding a study, if we do that trough our migrations, we know for sure that all required related models get created at the same time.

PR #216, opened 2026-04-02
feature: selectoptions can now be ordered.

### LINKED ISSUE #174: Determine order of SelectOptions
A simple `order_with_respect_to` should be fine, along with a modification in Django Admin similar to how questions are ordered within steps and steps are ordered within a form.

RoordinkFrank:
fixes #174 

Forms.py is slowly getting more duplicate code. I don´t think this can be easily changed but perhaps something to look into after the MVP.


RoordinkFrank:
I did just now notice load_fixtures does not work anymore. 
image

I am not sure if load_fixtures should be changed or the select option. Is is supposed to be nullable? I don´t think so?

backend/form/forms.py
```diff
@@ -1,8 +1,78 @@
 from django.forms import ModelForm, CharField, ValidationError, Textarea
-
 from form.models import Step
 
 
+class SelectQuestionAdminForm(ModelForm):
+    options_order = CharField(
```
XanderVertegaal: I prefer `option_order`, to keep things in line with `substep_order` and `question_order`. Make sure to also change this in `admin.py`.

backend/form/forms.py
```diff
@@ -1,8 +1,78 @@
...
+                continue
+            if not id_str.isdigit():
+                raise ValidationError(
+                    f"Invalid question ID format: '{id_str}'. Must be numeric."
+                )
+            requested_order.append(int(id_str))
+
+        # Validate that the provided IDs exist.
+        available_ids = [question.id for question in self.instance.options.all()]
+        # In instance.NAME.all(), NAME is the verbose name.
```
XanderVertegaal: This comment belongs to the line above it. I also think you mean _related_ name when you say _verbose_ name (which is something else).

XanderVertegaal (approved):
Works well!

As you mentioned, some parts of the code that are now duplicated (conversion of str -> list[int], making sure that all IDs exist) could be outfactored to reusable functions to make things a bit more DRY, but you may leave that for another PR if desired.

RoordinkFrank:
In VWR.json the order has been changed, likely because of questions and responses models being moved.

Only actual difference should be the _order field to questionoptions.

Means used to generate vwr.json 
1. python manage.py dumpdata --indent 4 form --exclude form.questionresponse --exclude form.userformsubmission > form/fixtures/vwr.json
2. Message xander to reformat the code. Usually works under right click on file -> reformat.

XanderVertegaal (approved):

PR #220, opened 2026-04-09
style: prettier

EdoStorm96:
I feel like, when I ran prettier, there were always unrelated files from before we had the prettier pr-check. This should fix that, I hope.


EdoStorm96:
Closing, because it is no longer up to date

Issue #191, opened 2026-03-18
Overview page and mark study as completed


XanderVertegaal:
At the end of the form, there should be a general (frontend-only) step outside of the form generated on the basis of the form config. This extra step should contain an overview of all entered answers and a big green 'Submit!' button.

Submission should only be possible if all steps are valid.

After submitting, a new StudyStatusChange should be created to set the Study to "SUBMITTED".

PR #224, opened 2026-04-14
Feature/overview step

XanderVertegaal:
Addresses part of #191 

This PR adds an additional page to a form that allows the user to go over their answers. The overview is not hardcoded onto a page but put into reusable components, so we can also show an overview on the Study detail page. 

image

Clicking the green button reveals an error if the form has any errors or if there are any unanswered required questions.

image

This is not very helpful if you don't know which questions are invalid. A different PR (#225) will take care of this.

~To avoid clashes with 'real' form steps, a new environmental variable "OVERVIEW_STEP_SLUG" is used in `compose.yaml`. I have also added a validator that runs whenever a Step is modified in Django Admin.~

Implementation-wise, a step now has a field `is_overview`. Steps with this property set to true will not show any attached questions but the overview of all answers. A validator is present to warn us if we attempt to add a question to an overview step, or if we mark a step with questions as overview. Lastly, the VWR fixtures have been updated to include this overview step.


frontend/components/form/overview/OverviewStep.vue
```diff
@@ -0,0 +1,50 @@
...
+const notAnswered = t("Not answered");
+
+function formatAnswer(question: QuestionWithValue): string {
+    const value = question.value;
+
+    switch (question.__typename) {
+        case "TrueFalseQuestionType":
+            return value ? t("Yes") : t("No");
+        case "FileUploadQuestionType":
+            return value instanceof File ? value.name : notAnswered;
```
XanderVertegaal: This is a bit of a placeholder for now. FileUploadQuestion is not really implemented yet.

XanderVertegaal:
Still do do: replace env var setup with adding step on the backend with an is_overview boolean. No route matching, no composable needed. The frontend can simply check the step and show an overview accordingly.

backend/form/fixtures/vwr.json
EdoStorm96, line 1: Could you also add the study_name_question to the `MRForm` fixture?

backend/form/models.py
```diff
@@ -72,6 +78,16 @@ class Meta:
             )
         ]
 
+    def clean(self) -> None:
+        """Validate that overview steps don't have questions."""
+        super().clean()
+        if self.is_overview and self.pk and self.questions.exists():  # type: ignore
```
EdoStorm96, line 111: I am not opposed to the idea of the overview step having questions ... So as far as I'm concerned you can remove this.

frontend/components/form/overview/OverviewForm.vue
```diff
@@ -0,0 +1,36 @@
...
+const contentSteps = computed(() =>
+    props.form.steps.filter((step) => !step.isOverview),
+);
+</script>
+
+<template>
+    <div v-for="step in contentSteps" :key="step.stepId" class="mb-4 p-2">
+        <h3>{{ useTranslateableAttribute(step, "name") }}</h3>
+
+        <div v-if="!step.substeps || step.substeps.length === 0">
```
EdoStorm96: I think we'll also want to show an OverviewStep for a step with substeps and questions. eg. datasubjecten has questions, but it also has substeps. Now the questions for the parent step are not shown.

frontend/components/form/overview/OverviewStep.vue
```diff
@@ -0,0 +1,59 @@
...
+
+function formatAnswer(question: QuestionWithValue): string {
+    const value = question.value;
+
+    switch (question.__typename) {
+        case "TrueFalseQuestionType":
+            return value ? t("Yes") : t("No");
+        case "FileUploadQuestionType":
+            return value instanceof File ? value.name : notAnswered.value;
+        case "SelectQuestionType":
```
EdoStorm96, line 37: selectquestions currently show the pk of the selectoption.

frontend/components/form/overview/OverviewStep.vue
```diff
@@ -0,0 +1,59 @@
...
+        case "TextQuestionType":
+        case "DateQuestionType":
+            return typeof value === "string" && value.trim()
```
EdoStorm96: Does trim remove newlines? Right now newlines are not preserved. I would prefer if they were, but it is not a dealbreaker.

frontend/components/form/FormWrapper.vue
```diff
@@ -95,6 +123,15 @@ const selectedStep = computed(() => {
     return steps.find(({ slug }) => slug === props.currentStepSlug) ?? null;
 });
 
+const firstStepSelected = computed(() => {
```
EdoStorm96, line 126: yay

frontend/pages/studies/[study_id].vue
```diff
@@ -102,7 +99,7 @@ const studyStatus = computed(() =>
                 </div>
             </div>
         </div>
-        <div class="uu-container" v-else>
+        <div v-else class="uu-container">
```
Comment (EdoStorm96) [inline frontend/pages/studies/[study_id].vue:102]: woops hehe 😅

frontend/Dockerfile.prod
```diff
@@ -17,7 +17,7 @@ COPY . /app
 
 # Get static graphql schema
 ARG CODEGEN_SCHEMA
-env CODEGEN_SCHEMA=${CODEGEN_SCHEMA}
+ENV CODEGEN_SCHEMA=${CODEGEN_SCHEMA}
```
EdoStorm96, line 20: What does this change exactly?

EdoStorm96 (approved):
Super nice feature! I found some minor things, but this is really nicely done, with not really a lot of new code, so it's great. I think you should address a few of my comment (displaying of selectquestion and parentsteps questions), and the rest I'll leave up to you. Nice feature.

frontend/Dockerfile.prod
XanderVertegaal: Nothing functional, but keywords in Dockerfiles are usually put in all-caps. This is just for consistency :)

backend/form/fixtures/vwr.json
XanderVertegaal: I'll leave that for another PR, as that feels like another issue. Also, with [this issue](https://github.com/CentreForDigitalHumanities/myresearch/issues/238) it seems like we're not sure yet how we want to implement Study names.

frontend/components/form/overview/OverviewForm.vue
XanderVertegaal: Right, good spot! 👍

frontend/components/form/overview/OverviewStep.vue
XanderVertegaal: No, this simply checks whether a value is not empty/falsy after removing spaces/whitelines. It's not supposed to change the value itself. I'll add a `!== ""` to make this more explicit.

XanderVertegaal: Whitespace is now preserved!

backend/form/models.py
XanderVertegaal: Hmm, I thought about this, but then I wasn't sure where those questions going to be 'overviewed' themselves. It would be weird to have questions on an overview page being part of the overview shown on that same page.

If we reuse this component on the Study detail page, it **does** make sense to display those overview step questions there, so maybe we need a bit of filtering.

PR #225, opened 2026-04-14
Feature/show errors in overview

### LINKED ISSUE #204: Validate before finalizing submission
On the Study detail page, two buttons:
- Continue (to the first step of the form)
- Submit (to the overview page of the form)

On the overview page, there should be a Submit button that is only clickable if the form is valid.

XanderVertegaal:
Ready for review!


frontend/composables/useFormErrors.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        // Add substeps to the stack if they exist.
+        if ("substeps" in currentStep && currentStep.substeps) {
+            stepsToCheck.push(...currentStep.substeps);
+        }
+    }
+
+    return false;
+}
```
miggol: Interesting! I can see how this is logical if we really need a way to check a single step. But there is now a lot of re-checking of the same questions going on on the overview page.

A bottom-up approach from the question level would be more efficient. Turn the tree upside down and if you encounter a question with an error, you can immediately flag all its parent steps as having an error.

Then we could store that info on the step and question objects so it can be reused. Would be useful to avoid even more checking if we want to give the stepper error markers.

Think about it if you fancy a programming challenge, lol. Otherwise this works great so totally not blocking, don't worry about it :)

miggol (approved):
Clean! And the end product looks really nice as well.

frontend/composables/useFormErrors.ts
XanderVertegaal: Ah, I see your point! It would indeed be neater if we didn't have to traverse the form so often. In the end I could not easily flip the tree and work bottom-up (although I support the idea!). However, I did manage to get rid of the multiple traversal problem by adding a function `useAnnotateErrors` that marks errors on questions in steps. It traverses the form once and marks errors and invalid statuses on the questions and steps themselves. The form object then holds all relevant error information and we don't need pass the validation object to various child components anymore. Hopefully, this makes it easier for the Stepper to display error markers as well :)

XanderVertegaal: Okay, I played with it some more and I am pretty confident we can hide simplify things even more by building the form object and annotating it with errors in one go. That would mean we can hide more validation logic in `useProcessForm.ts` (where a lot of the business logic happens anyway) and away from `FormWrapper.vue` (so it can be a dumb display component). This is not something for this PR, but more "licht aan de horizon" 😉

PR #227, opened 2026-04-15
hidden form values

### LINKED ISSUE #176: Hidden form fields with values
Questions that are hidden because of a condition 'remember' the values the user entered previously. This is a problem if a user first enters an invalid value into field A and then changes field B in such a way that field A is hidden. The form as a whole will be invalid, but the user can't see the offending form field.

Can we simply remove QuestionResponses for questions that are not visible after passing through the FormResolver?

RoordinkFrank:
fixes #176

- Does address repeating questions but does not address repeating steps. I did not see anything in the issue about that. Will those be a problem as well?
- as a heads up: I am not sure if I was doing something wrong but in an older commit version of this code the incoming repeating question_responses often had the same repeat_index.  I have not yet been able to replicate it with the newer code setup so hopefully it's all good. I assume for now there was something wrong in the old code even though i triple checked that as well. Perhaps I made a mistake in the db somewhere. In any case if  we ever have repeating_questions that are not deleted properly then it may be a good idea to look at the repeat_index of the question_responses. I was spamming volgende and vorige so that may be a clue.
- As a general note, the form volgende, vorige and any other navigational option should really be greyed out when the form is reloading, such as making an invis question visible. If you go volgende while this is happening and go back you have suddenly 2 repeating questions while you should have 3 etc.

--------------------
VWR does not have repeating questions right now so they have to be manually added.
Used setup:

image

image

image


backend/form/services/user_form_resolver.py
```diff
@@ -115,13 +115,29 @@ def _resolve_step_instances(self, step: Step) -> list[StepType]:
         return instances
 
     def _resolve_question_instances(self, question: BaseQuestion) -> list:
-        """Resolve all instances of a question (considering repeats)."""
+        """Resolve all instances of a question (considering repeats).
+        Also, deletes hidden response."""
```
XanderVertegaal:
```suggestion
        Also deletes hidden responses as a side-effect."""
```

backend/form/services/user_form_resolver.py
```diff
@@ -115,13 +115,29 @@ def _resolve_step_instances(self, step: Step) -> list[StepType]:
...
+
+        # non-repeating questions have a repeat count as well
+        repeat_count = self.evaluator.get_repeat_count_for_question(question)
+
+        response_object = QuestionResponse.objects.filter(
```
XanderVertegaal: I prefer `response_objects` or `responses`, since `.filter()` produces a queryset.

backend/form/services/user_form_resolver.py
```diff
@@ -115,13 +115,29 @@ def _resolve_step_instances(self, step: Step) -> list[StepType]:
...
+        # form filled in wrong errors triggering on invisible questions.
         if not self.evaluator.is_question_visible(question):
+            response_object.delete()
             return []
 
-        repeat_count = self.evaluator.get_repeat_count_for_question(question)
-        instances = []
+        # With repeating questions only the responses over the repeat_count
+        # are considered hidden so only those responses need to go
+        if response_object.count() > repeat_count:
```
XanderVertegaal: This condition is unnecessary.

backend/form/services/user_form_resolver.py
```diff
@@ -115,13 +115,29 @@ def _resolve_step_instances(self, step: Step) -> list[StepType]:
...
+        # non-repeating questions have a repeat count as well
+        repeat_count = self.evaluator.get_repeat_count_for_question(question)
+
+        response_object = QuestionResponse.objects.filter(
+            submissions__in=[self.evaluator.submission],
+            question=question,
+        )
+
+        # If a response is hidden it should be removed to prevent possible
+        # form filled in wrong errors triggering on invisible questions.
```
XanderVertegaal:
```suggestion
        # Hidden responses should be removed to prevent invisible form errors.
```

XanderVertegaal (commented):
I didn't test this, but the code looks good to me. My comments are mainly about the documentation, so nothing major. If you make the suggested changes (and remove the unnecessary `if`) you can consider this approved.

EdoStorm96 (approved):
Nice and clean! Code looks good and I've tested this and it works well.

I guess refreshing repeating questions need a little work (as you probably noticed), as it seems pretty buggy, but that is unrelated to this PR. Good work :)

PR #233, opened 2026-04-21
Feature/study actions

### LINKED ISSUE #196: Study actions
Current suggestions

* Continue editing
* Delete study
* Overview (all on page read-only)
* Submit
* Unsubmit

An existing implementation of this can be found in Ethics, the ReviewActions.

EdoStorm96:
fixes #196

Was pretty easy in the end. The backend is pretty much a copy of ReviewActions from the FETC portal, except we just send some strings to the frontend, which get defined in a `StudyAction` object. This is also where the logic for making a specific action available lives.

These strings are then mapped in the frontend to an object which contains a bit more info on where to send everything. I've also added the possibility to add some additional styling to an action.

For now I've implemented the continue action using this system and I have a delete action which links to a placeholder page.

I've designed these action to redirect to a page where you can do a mutation (or something) as opposed to being able to do mutations directly from the study dashboard. I think this is fine and keeps our dashboard clean.

I don't really love that the strings which are our keys are hardcoded in the frontend. Any tips to get around this?

All fake action tiles are now removed.

I've also renamed some paths to use `[camelCase]` instead of `[snake_case]` which makes this diff more messy then it should be ... But I had to do this to make navigateTo work nicely.

Most important files to check out are `study_actions.py` and `availableActions.vue`

little impression (the danger color is on hover):
image


frontend/pages/studies/index.vue
EdoStorm96, line 1: Looks like I ran prettier here.

backend/research/utils/study_actions.py
```diff
@@ -0,0 +1,88 @@
...
+#################
+
+
+class EditAction(StudyAction):
+
+    action = StudyActionEnum.EDIT_ACTION
+
+    def is_available(
+        self,
+    ):
```
XanderVertegaal: (For ease of reading:)

```suggestion
    def is_available(self):
```

backend/research/utils/study_actions.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if (
+            self.study in Study.objects.accessible_objects(self.user, MRPermission.EDIT)
+            and self.study.status.status == SubmissionStatus.DRAFT
+        ):
+            return True
+
+        return False
```
XanderVertegaal:
```suggestion
        return (
            self.study in Study.objects.accessible_objects(self.user, MRPermission.EDIT)
            and self.study.status.status == SubmissionStatus.DRAFT
        )
```

backend/research/utils/study_actions.py
```diff
@@ -0,0 +1,88 @@
...
+        self.study = study
+        self.user = user
+
+    def is_available(
+        self,
+    ) -> bool:
+        """Returns true if this action is available to the specified
+        user given the current review."""
+
+        return True
```
XanderVertegaal, line 52: If you forget to implement an action, it automatically gives the permission. Don't you want to return `False` here instead or `raise NotImplementedError`?

backend/research/utils/study_actions.py
```diff
@@ -0,0 +1,88 @@
...
+
+class StudyActions:
+    """
+    An object which can evaluate the available actions for a study and a user.
+    """
+
+    def __init__(self, study: Study, user: User):
+        self.study = study
+        self.user = user
+        self.all_actions = [EditAction(study, user), DeleteAction(study, user)]
```
XanderVertegaal: Does this mean we have to remember to manually add actions here after defining them if we want them to be used?

frontend/pages/studies/[studyId]/delete.vue
Comment (XanderVertegaal) [inline frontend/pages/studies/[studyId]/delete.vue:1]: I propose we use a "Are you sure?" modal for this. In that case we don't really need a page here (although it doesn't hurt either).

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,103 +1,102 @@
...
-    {
-        label: t("Continue editing"),
-        href: continueUrl.value,
-        icon: PencilLine,
-    },
-    {
-        label: t("View attachments"),
-        href: "#",
-        icon: Paperclip,
+// NOTE: ensure that the strings used as keys here correspond with
```
XanderVertegaal: This comment can go now! 🥳

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,103 +1,102 @@
...
+      <NuxtLink
+        v-for="(action, index) in availableActions"
+        :key="index"
+        @click.prevent="handleActionClick(action)"
+        :style="action.style"
+        class="tile h-100 justify-content-around"
+      >
+        <strong class="text-center">{{ $t(action.label) }}</strong>
+        <component v-if="action.icon" :is="action.icon"> </component>
+      </NuxtLink>
```
XanderVertegaal: Another silly/small convention thing with no functional difference: event handlers like `@click` are normally put all the way at the end. (My IDE/ESLint complains about this.)


```suggestion
            <NuxtLink
                v-for="(action, index) in availableActions"
                :key="index"
                :style="action.style"
                class="tile h-100 justify-content-around"
                @click.prevent="handleActionClick(action)"
            >
```

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,103 +1,102 @@
...
 
 type AvailableAction = {
-    label: string;
-    href: string;
-    icon: Component;
+  label: string;
+  name: string;
+  params: Record<string, any>;
+  icon?: Component;
+  style?: Record<string, string>;
```
XanderVertegaal: If you don't know the type and don't care what it is exactly, you can use `unknown` as a safer alternative to `any`.

frontend/pages/studies/[studyId]/index.vue
```diff
@@ -0,0 +1,110 @@
+<script lang="ts" setup>
+import { useQuery } from "@vue/apollo-composable";
+import { graphql } from "~/generated/gql";
+import type { GetStudyQuery } from "~/generated/gql/graphql";
+import StudyDetailsSidebar from "../../../components/study_detail/StudyDetailsSidebar.vue";
+import AvailableActions from "../../../components/study_detail/AvailableActions.vue";
+import StudyProgessBar from "../../../components/study_detail/StudyProgessBar.vue";
+
+const GET_STUDY = graphql(`
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[studyId]/index.vue:9]: It feels more logical to fetch the actions in the AvailableActions component itself, since you're not really using them here. You can add the study-actions query to the query you're already performing in that component, so you wouldn't be adding an extra query.


(In AvailableActions.vue:)
```ts
const GET_FIRST_SLUG_AND_ACTIONS = graphql(`
    query GetFirstSlugAndActions($studyId: ID!, $submissionId: ID!) {
        form(submissionId: $submissionId, mrPermission: "Edit") {
            formId
            steps {
                stepId
                slug
            }
        }
        study(id: $studyId) {
            id
            actions
        }
    }
`);
```

frontend/pages/studies/index.vue
XanderVertegaal: If I run Prettier, all changes go away (apart from the one in NuxtLink). Maybe we're not running the same version of Prettier?

backend/research/utils/study_actions.py
```diff
@@ -0,0 +1,88 @@
...
+
+        return True
+
+
+#################
+# Study actions #
+#################
+
+
+class EditAction(StudyAction):
```
XanderVertegaal: If this file is going to have actions for more than one model, maybe `StudyEditAction` is safer to avoid future clashes.

backend/research/utils/study_actions.py
```diff
@@ -0,0 +1,88 @@
+from graphene import Enum
+
+from research.other_models.reviews import SubmissionStatus
+from main.models import MRPermission, User
+from research.models import Study
+
+
+class StudyActionEnum(Enum):
```
XanderVertegaal: Maybe we can reuse this for other objects as well, in which case "Study" unnecessarily limits the scope of this enum. In that case we could call it `ActionEnum`, or maybe even just `Action`.

backend/research/utils/study_actions.py
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self, study: Study, user: User):
+        self.study = study
+        self.user = user
+        self.all_actions = [EditAction(study, user), DeleteAction(study, user)]
+
+    def get_available_actions(self) -> list[StudyActionEnum]:
+        return [a.action for a in self.all_actions if a.is_available()]
+
+
+class StudyAction:
```
XanderVertegaal: This looks like a good case for an [Abstract Base Class](https://pythonguides.com/python-abstract-base-classes-abc/). It makes it impossible to instantiate this class directly and you can use it to enforce that every subclass defines an `action`, like so:

```py
from abc import ABC, abstractmethod

class StudyAction(ABC):
    
    @property
    @abstractmethod
    def action(self):
        pass

```

If I create a new study action subclass without defining `action`, I will get a build time (if that's a thing in Python) error instead of a runtime error.

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,103 +1,102 @@
...
 });
 
-const continueUrl = computed(() =>
-    slug.value ? `/procreg/${props.submissionId}/${slug.value}` : "",
-);
+const isSlugLoaded = computed(() => slug.value !== null);
 
 const { t } = useI18n();
 
 type AvailableAction = {
```
XanderVertegaal: Please move types and interface all the way to the top of the file. (That's where people will expect to find them.)

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,103 +1,102 @@
...
+    label: t("Continue editing"),
+    name: "procreg-submissionId-slug",
+    params: { submissionId: props.submissionId, slug: slug.value },
+    icon: PencilLine,
+  },
+  [StudyActionEnum.DeleteAction]: {
+    label: t("Delete"),
+    name: "studies-studyId-delete",
+    params: { studyId: props.studyId },
+    icon: Delete,
```
XanderVertegaal: Hmm, this icon calls to mind more a 'delete character in a text box' idea, rather than a 'delete object' idea. How about the good ol' trash bin?

frontend/components/study_detail/AvailableActions.vue
XanderVertegaal: Another ESLint warning: `Attribute ":is" should go before "v-if".`

XanderVertegaal (changes requested):
Well done, this works really well! I am happy to see that we can avoid string matching and that the frontend takes care of all the display stuff (strings + icons). It's very cleanly separated this way.

Apart from a set of smallish nitpicks which you may ignore as desired, the only things I would encourage you to consider are:
- Turn StudyAction into an ABC (if you think it's appropriate).
- Move the actions query into the AvailableActions component.
- Replace `any` with `unknown`.

Also, see if you can enable ESLint in your IDE to spot things like `attribute :is should go before v-if` and the indentation issue, or at least set your IDE to use 4 spaces as the default indent to avoid the linting issues.

backend/research/utils/study_actions.py
EdoStorm96: Yes ...

backend/research/utils/study_actions.py
EdoStorm96: I guess not really necessary if we go the abc route?

frontend/pages/studies/[studyId]/delete.vue
Comment (EdoStorm96) [reply frontend/pages/studies/[studyId]/delete.vue:1]: Let's do this in a future PR

frontend/pages/procreg/[submissionId]/[slug].vue
```diff
@@ -116,7 +116,7 @@ const GET_FORM = graphql(`
 const route = useRoute();
 
 const submissionId = computed(() => {
```
Comment (XanderVertegaal) [inline frontend/pages/procreg/[submissionId]/[slug].vue:118]: My URL change PR (#235) has a nice little composable useSubmissionId that we can use for this (and elsewhere). I'll make sure to implement that here after this PR is merged! :)

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,103 +1,112 @@
...
+                :style="action.style"
+                class="tile h-100 justify-content-around"
+                @click.prevent="handleActionClick(action)"
+            >
+                <strong class="text-center">{{ $t(action.label) }}</strong>
+                <component :is="action.icon" v-if="action.icon"> </component>
+            </NuxtLink>
+        </div>
     </div>
+    <h3 v-else>{{ $t("No actions available") }}</h3>
```
XanderVertegaal, line 111: Nice addition of this 'empty text' and a loading spinner! 👍

frontend/pages/studies/[studyId]/index.vue
```diff
@@ -2,9 +2,9 @@
 import { useQuery } from "@vue/apollo-composable";
 import { graphql } from "~/generated/gql";
 import type { GetStudyQuery } from "~/generated/gql/graphql";
-import StudyDetailsSidebar from "../../components/study_detail/StudyDetailsSidebar.vue";
-import AvailableActions from "../../components/study_detail/AvailableActions.vue";
-import StudyProgessBar from "../../components/study_detail/StudyProgessBar.vue";
+import StudyDetailsSidebar from "../../../components/study_detail/StudyDetailsSidebar.vue";
+import AvailableActions from "../../../components/study_detail/AvailableActions.vue";
+import StudyProgessBar from "../../../components/study_detail/StudyProgessBar.vue";
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[studyId]/index.vue]: You can use `~` as a shorthand for the frontend root directory :)

```suggestion
import StudyDetailsSidebar from "~/components/study_detail/StudyDetailsSidebar.vue";
import AvailableActions from "~/components/study_detail/AvailableActions.vue";
import StudyProgessBar from "~/components/study_detail/StudyProgessBar.vue";
```

XanderVertegaal (approved):
Beautiful! <3

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -1,103 +1,112 @@
...
+}));
 
-const availableActions = computed(() =>
-    props.studyStatus === "draft" ? draftActions.value : POActions.value,
-);
+const availableActions = computed<AvailableAction[]>(() => {
+    const studyActions = result.value?.study?.actions;
+    if (studyActions) {
+        return studyActions.map(
+            (actionEnum: ActionEnum) => actionMap.value[actionEnum],
```
XanderVertegaal, line 85: (I don't think you should have to type actionEnum explicitly. Your IDE should know what it is.)

PR #234, opened 2026-04-21
Fix Vue-TSC watcher (and codegen) in frontend container

### LINKED ISSUE #231: Fix Vue-TSC and codegen watcher in dev container
Template and code changes are picked up by Vite (live reload) but the TS type checker (Vue-TSC) and GraphQL codegen do not seem to pick up changes. Currently, restarting the frontend container after every code change is an unpleasant workaround.

XanderVertegaal:
Fixes #231 

With the kind help of @EdoStorm96 I've recently switched to using the Docker containers for development. (I used to run local dev servers on my machine before.)

I noticed that our current setup picks up code changes correctly but does not run the Vue typechecker (`vue-tsc`). As a result, type errors that are there when the container is started never go away, and new type errors are not picked up until you manually restart the container, which is quite annoying.

(I also seem to have had at least one case in which GraphQL codegen failed to run when I made a change to one of the queries, but I've heard from @EdoStorm96 that codegen does work on his machine, so maybe that was a fluke.)

This PR implements `docker compose watch` using [this guide](https://docs.docker.com/guides/vuejs/develop). By running `docker compose --profile dev up --watch` you start the containers in watch mode. For the `vue-dev` service, this means that code changes are continuously synced (and a rebuild is triggered is dependencies are added). This means we don't need the volumes anymore.

To test, run `docker compose --profile dev up --watch` and:
- Update a template. (This already works on develop.)
- Commit a type error somewhere (e.g. use an unknown function) and verify that the type error pops up in the browser.
- Update a query and verify that the types are regenerated (by codegen).

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -123,17 +123,27 @@ services:
       - "3000"
     ports:
       - "3000:3000"
-    volumes:
-      # Enable live-reloading of frontend code.
-      - ./frontend:/app
-      # Use container's node_modules.
-      - frontend-node-modules:/app/node_modules
     environment:
       # Allows Nuxt to be accessible from the outside.
       - HOST=0.0.0.0
       - CODEGEN_SCHEMA=http://mr-django:8000/api/graphql
       - NUXT_PUBLIC_API_URL=${BASE_URL}/api
       - NUXT_PUBLIC_SAML_URL=${BASE_URL}/saml
+    develop:
+      watch:
+        - path: ./frontend
+          target: /app
+          action: sync
+          ignore:
+            - node_modules
+            - .nuxt
+            - dist
+        - path: ./frontend/package.json
+          target: /app/package.json
+          action: restart
+        - path: ./frontend/vite.config.js
+          target: /app/vite.config.js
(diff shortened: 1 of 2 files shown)


compose.yaml
```diff
@@ -123,17 +123,27 @@ services:
...
+        - path: ./frontend
+          target: /app
+          action: sync
+          ignore:
+            - node_modules
+            - .nuxt
+            - dist
+        - path: ./frontend/package.json
+          target: /app/package.json
+          action: restart
```
miggol, line 143: Wouldn't this need to be a rebuild, say, if we were to add a dependency to `package.json`?

miggol (approved):
Only slightly embarrassing us container users just took these errors for granted all this time.

I have one question but feel free to merge afterwards!

compose.yaml
XanderVertegaal: Good point. I guess it would, but in our case the entrypoint script runs `npm install` when starting the container, so new dependencies should be pulled in on every `up`.

PR #235, opened 2026-04-21
Update URL structure

### LINKED ISSUE #228: change url for form
Change the url for the form from `procreg/[submissionId]/[slug]` to `studies/[studyId]/[submissionId]/[slug]`. Would be nice imo.

XanderVertegaal:
Closes #228

Moves the start page to `/start` and the form pages to `/studies/studyId/submissionId/slug`.


frontend/components/study_detail/AvailableActions.vue
```diff
@@ -90,14 +101,14 @@ const availableActions = computed(() =>
 <template>
     <h3 class="mb-3">{{ $t("Available actions") }}:</h3>
     <div class="tiles">
-        <a
+        <NuxtLink
             v-for="(action, index) in availableActions"
             :key="index"
-            :href="action.href"
+            :to="action.href"
```
XanderVertegaal: This is a minor, unrelated optimisation. Hrefs/anchor tags are used for links to external pages, so currently this triggers a full page reload. With `NuxtLink`, you stay within the SPA, making it a bit more smooth.

frontend/composables/useRouteParams.ts
XanderVertegaal, line 1: A set of handy utils that I feel will come in use later. For some reason, the type checker here (unlike DIAPP's) thinks that  `route.params.study_id`, `route.params.submission_id` etc. are always of type `string | string[]`, which isn't true. If I call it on a route without a route param, it yields `undefined`.

I'm marking it here as potentially undefined, which is a bit 'roomser dan de paus' and annoying to deal with, but more type-safe.

frontend/pages/studies/[study_id]/index.vue
```diff
@@ -94,15 +99,15 @@ const studyStatus = computed(() =>
...
+                        <!-- Progress bar -->
                         <div class="col-2">
                             <StudyProgessBar :study-status="studyStatus" />
                         </div>
                     </div>
                 </div>
             </div>
         </div>
-        <div class="uu-container" v-else>
+        <div v-else class="uu-container">
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[study_id]/index.vue:108]: Structural directives like `v-if`, `v-for` etc. always come before all other attributes.

frontend/components/study_detail/AvailableActions.vue
```diff
@@ -33,9 +34,19 @@ const slug = computed<string | null>(() => {
...
+const continueUrl = computed(() => {
+    const formSlug = slug.value;
+    const submissionId = props.submissionId;
+    const studyIdValue = studyId.value;
+
+    // TODO: disable/hide the button if the slug is not available (yet).
+    if (!formSlug || !submissionId || !studyIdValue) {
+        return "#";
+    }
+    return `/studies/${studyIdValue}/${submissionId}/${formSlug}`;
```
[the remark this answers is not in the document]
XanderVertegaal: This will be fixed in a future PR.

frontend/components/shared/FormWrapper.vue
```diff
@@ -154,8 +157,13 @@ function getPreviousStepSlug(): string {
...
 function navigateToSlug(slug: string) {
     submitForm();
+
+    const id = studyId.value;
+    if (!id) {
+        return;
+    }
     return navigateTo(
-        "/procreg/" + props.queriedForm.submissionId + "/" + slug,
+        `/studies/${id}/${props.queriedForm.submissionId}/${slug}`,
```
EdoStorm96: In #233 I've mostly switch to the name & params syntax for navigateTo. I think that is a bit nicer
```suggestion
        {name: "studies-studyId-submissionId-slug", params: {studyId: id, submissionId: props.queriesForm.submissionId, slug: slug}}
```
This suggestion is not really ready to commit though (besides formatting), as you would need to rename the folders to use camelCase i/o snake_case. I think this is the direction we should be going, with naming [slug] folders and pages, as camelCase is the most flexible in my experience.

frontend/components/study_detail/AvailableActions.vue
EdoStorm96, line 1: We'll have some merge issues when this gets merged with #233 ... But we'll fix that in the order things get merged.

EdoStorm96 (changes requested):
While I think this is a good idea, I would prefer folder names to be camelCase i/o snake_case (even though I started doing it in snake_case ... It turns out this is not ideal ... Could you change it to this?

With AvailableActions, there will be some merge issues with #233, but we'll cross that bridge when we get there ...

PR #236, opened 2026-04-22
Split up form models into individual files

miggol:
I did this at the start of my new feature but you guys are developing way too fast, causing me to have to de a lengthy rebase rather often. Hope you all agree this tidies up the `models.py` situation :)

Form tests get the same treatment. Should be functionally equivalent.


backend/form/models/questions.py
```diff
@@ -0,0 +1,84 @@
+from django.db import models
+
+
+# Questions
```
XanderVertegaal: These comments that are now at the top of each file were meant to give structure to our old monolithic `models.py`. As far as I am concerned, these can go now, but I'll leave that up to you :)

XanderVertegaal (approved):
Great, much cleaner :)

PR #247, opened 2026-05-11
Add reusable modals

XanderVertegaal:
This PR introduces reusable modals that can be created using the `useConfirm` composable in `cdh-vue-lib`.

image

This PR is open for review, but it should not be merged until https://github.com/CentreForDigitalHumanities/Vue-lib/pull/16 is merged. (This branch currently refers to a specific commit on `cdh-vue-lib`, not a specific release.) As such, the implementation of `useConfirm` in `cdh-vue-lib` is still a bit 'experimental' and open to change.


frontend/components/studyDetail/StudyProgessBar.vue
XanderVertegaal, line 1: No functional changes; just trying to get Prettier to pass.

XanderVertegaal (commented):
I've made a pre-release of VueLib 0.7.0 and am now referencing that version in `package.json` so we can go ahead and review this.

EdoStorm96 (approved):
Yes, this is super nice! Awesome to have this functionality and a nice example! Would be nice in the future if we can do some styling, but that's for another time.

PR #248, opened 2026-05-11
fix: simple fix for writing to generated outside container

EdoStorm96:
We've just noticed that since #234 has been merged, our codegen generated files no longer updates outside of the container. @XanderVertegaal helped me with a fix for this. This does solve the problem, but I'm not sure if having the user be root is ideal/necessary. I ran into permission problems otherwise. @miggol, do you have thoughts on this?


compose.yaml
```diff
@@ -129,6 +129,9 @@ services:
       - CODEGEN_SCHEMA=http://mr-django:8000/api/graphql
       - NUXT_PUBLIC_API_URL=${BASE_URL}/api
       - NUXT_PUBLIC_SAML_URL=${BASE_URL}/saml
+    volumes:
+      - ./frontend/generated:/app/generated
+    user: root
```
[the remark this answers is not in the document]
XanderVertegaal: Using `user: "${UID}:${GID}"` seems to work on my Windows machine and may be better than `root`, although this should never run on production, of course.

compose.yaml
```diff
@@ -129,6 +129,9 @@ services:
       - CODEGEN_SCHEMA=http://mr-django:8000/api/graphql
       - NUXT_PUBLIC_API_URL=${BASE_URL}/api
       - NUXT_PUBLIC_SAML_URL=${BASE_URL}/saml
+    volumes:
+      - ./frontend/generated:/app/generated
+    user: "${UID}:${GID}"
```
miggol: If this works, it would be nice to have it on the django-dev service as well. Saves having to fix permissions after running `makemigrations` and such.

frontend/Dockerfile.dev
```diff
@@ -20,12 +20,6 @@ RUN --mount=type=cache,target=/root/.npm npm install
 
 COPY . .
 
-# Change ownership of the app directory to the node user.
-# This is necessary for watch mode.
-RUN chown -R node:node /app
-
-USER node
-
```
miggol, line 28: Without these lines, my vue-dev container can't access things in /app/, because inside the container it's still running as the node user either way.

miggol (commented):
Can't get this to work right now, will try again later!

frontend/Dockerfile.dev
miggol: This turned out to be unrelated to my issues

compose.yaml
miggol: `vue-dev` actually won't work running as anything other than root. So let's just drop the less-privileged user idea for now. The volume appears to fix access issues

miggol (changes requested):
I suggest removing the `user: "${UID}:${GID}"` attribute, as at least on linux running with an unprivileged UID breaks the vue-dev container. And without specifying those vars in `.env` it changes nothing.

Once again on linux (not sure about win-docker) setting a user in the Dockerfile has no effect outside of the container, if Docker itself runs as root. In such cases, files edited by the container are still edited by root, unless a custom `user:` attribute is set. Although this does affect permissions inside the container, gains are marginal.

So feel free to do what is needed to fix codegen on Windows. For me, codegen runs fine on this branch as long as both $UID and $GID are unset, i.e. as if the `user:` attribute did not exist.

Hopefully, besides adding a volume, removing `USER node` from the Dockerfile is all that was necessary to fix codegen on Windows.

XanderVertegaal (approved):
I can confirm that it works with just the volume and removing `USER node` from the Dockerfile.

(Adding a field `hello` to `StudyType` and then requesting that field in a frontend query neatly shows the new field in `frontend/generated/gql/graphql.ts`



I'll remove the `user` line from `compose.yaml` and this is good to be merged afterwards, as far as I am concerned.

PR #249, opened 2026-05-11
feat: working dates in study form

### LINKED ISSUE #186: Dates
Our stakeholders do not appreciate the representation of dates in non-dutch locales. Might be tricky to fix, but something to look into.

### LINKED ISSUE #226: Use "real" dates on study detail page
Currently on the study page, we display two dates which are randomly generated. They are now "Created on:" and "Submitted on:". I'd like to change "Submitted on" to the `UserFormSubmission`'s (USF) `updated_at`. so the dates displayed will be "Created on:", which displays the study's `created_at` attribute and "Updated on:", which displays the `updated_at` of the study's most recent USF.

To make this work, we'll need to make some other changes to the codebase.

1. Our first issue is that USF's `updated_at` is currently now working. I would suggest to update this manually in update_submission.py. This should get updated to `timezone.now()` when:

    - a new response gets added to a USF
    - an existing response's answer gets updated.

    Do this using this syntax:  `current_submission.update(updated_at=timezone.now())`

2. Then we'll need to the study's created_at to the `StudyType`, and add a custom field to the `StudyType` to retrieve the latest USF's updated_at date.

3. We'll then pass these on the the frontend via our `getStudy` query and display the new dates in our sidebar.

4. I'd like to have them formatted as `YYYY-MM-DD, HH:mm` for now (eg. 2026-04-14, 11:56).

5. Then we can remove the random date generator from our code.

Optionally, remove `auto_now=true` from `UserFormSubmission.updated_at`, as it is not really fit for our purpose ...

RoordinkFrank:
fixes #226

Edit: also fixes #186


backend/form/services/update_submission.py
```diff
@@ -9,9 +10,9 @@ def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSub
     # That is why we use .get() here, which handles both cases.
     try:
         # first filter for editable objects, and then try to get the specific submission
-        current_submission = UserFormSubmission.objects.accessible_objects(
-            user, MRPermission.EDIT
-        ).get(
```
RoordinkFrank, line 14: My Editor cann´t find the type automatically without setting the type explicit. That or I have some setting wrong.

frontend/composables/useLocalisation.ts
```diff
@@ -20,3 +20,10 @@ function useTranslateableAttribute<Key extends string>(
 }
```
RoordinkFrank, line 20: renamed to localisation to better reflect what is going on.

backend/form/services/update_submission.py
```diff
@@ -50,5 +51,6 @@ def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSub
                 repeat_index=response["repeat_index"],
             )
             new_response.submissions.add(current_submission)
-
+    current_submission.updated_at = timezone.now()
```
EdoStorm96: Now you always change the updated_at attribute, even if no new response.answer has been saved. Ideally, you would only update the updated_at if a new answer has been saved. I know this logic is pretty annoying to read (sorry ...). The updated_at should really only be updated if either:
- the if-clause on line 33 gets hit (yes, this is for revisions, which we do not yet support, but we will in the future, so best to keep this code up-to-date)
- the else-clause on line 43 gets hit.
- the else-clause on line 47 gets hit.

I think your best strategy is to have a variable in all these cases. Now you already have the new response variable to check for two of these case. I would suggest using this, but naming it new_answer or something. You can define this before the for loop as `new_answer = False`, then update it within the for-loop as we are using new_response, but be sure to also update it within the else-clause on line 43. Then after the forloop, `if new_answer:` update update_at.

I hope I made this clear haha. Let me know otherwise :)

EdoStorm96 (changes requested):
Nice work! I have a small comment, but I think mostly this looks quite good :) If you could address my comment, I'll approve.

backend/research/models.py
```diff
@@ -27,6 +25,9 @@ class Study(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
     created_by = models.ForeignKey(User, on_delete=models.CASCADE)
 
+    def updated_at(self) -> datetime:
```
XanderVertegaal: Since this method only takes `self`, can it be decorated with @property? If so, I'd like it to be moved below to the other properties. If not, I'd like it to be moved even further down. Ordinary class methods are usually put below properties (and Meta).

frontend/composables/useLocalisation.ts
```diff
@@ -20,3 +20,10 @@ function useTranslateableAttribute<Key extends string>(
...
 
 export { useTranslateableAttribute };
+
+const timeFormat: Intl.DateTimeFormatOptions = { month: 'long', day: 'numeric', year: 'numeric', hour: 'numeric', minute: 'numeric'}
```
XanderVertegaal: Nice that you extracted this and explicitly typed it with `Intl.DateTimeFormatOptions`! 👍

XanderVertegaal: You will probably have to run Prettier here to make sure everything is formatted nicely. I'd also like a whiteline in between `const timeFormat` and `function localDateTime`.

frontend/composables/useLocalisation.ts
```diff
@@ -20,3 +20,10 @@ function useTranslateableAttribute<Key extends string>(
...
+function localDateTime(isoString: string) {
```
XanderVertegaal: To make clear that this is a composable function, please call it `useLocalDateTime`. It doesn't change anything, but it's a conventional thing so we know we don't have to look for this function in a component `<script>` tag. ;)

frontend/composables/useLocalisation.ts
```diff
@@ -20,3 +20,10 @@ function useTranslateableAttribute<Key extends string>(
...
+    // localizes the isoString (from for example UTC time) into local datetime and handles translation
+    return new Date(isoString).toLocaleTimeString(i18n.global.locale.value, timeFormat)
+}
+export { localDateTime };
```
XanderVertegaal: There is another `export` statement elsewhere in this file. Please merge them and put both of them at the end of the file so there is only one `export` statement that exports both functions.

We started using this explicit `export { ... }` syntax for composables in DIAPP, but I'm no longer a fan of it: it's more conventional these days to simply use `export function useLocalDateTime(...)` (without a separate line with `export { ... }`). If you want, you can use that here as well, but if you do, please add `export` to `function useTranslateableAttribute` as well ;)

frontend/pages/studies/[studyId]/index.vue
```diff
@@ -34,24 +36,6 @@ const { result: studyResult } = useQuery<GetStudyQuery>(
 
 const study = computed(() => studyResult.value?.study ?? null);
 
-// Some functions to generate mockdata
-
-function randomDatePastYear(): string {
-    const today = new Date();
-    const oneYearAgo = new Date();
-    oneYearAgo.setFullYear(today.getFullYear() - 1);
-
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[studyId]/index.vue:43]: I'm glad that this can go!

frontend/components/studyDetail/StudyDetailsSidebar.vue
```diff
@@ -1,10 +1,14 @@
...
 
-defineProps<{
+interface Props {
     study: NonNullable<GetStudyQuery["study"]>;
-    randomDatePastYear: string;
-}>();
+}
+const props = defineProps<Props>();
+
+const createdAt = computed(() => localDateTime(props.study.createdAt));
```
XanderVertegaal: ESLint complains here: `Unsafe argument of type any assigned to a parameter of type string.`

The official GraphQL spec does not have a `DateTime` scalar (although there is a [work in progress here](https://www.graphql-scalars.com/date-time/)). Graphene adds one for you so you can work with it in Python, but GraphQL Codegen, which generates your TS types, only works with the official spec. As a result, `DateTime` is typed as `any`.

In our case, our 'datetimes' will probably only be ISO-compliant `strings`, so we can tell GraphQL Codegen that any field with the `DateTime` type will be a `string` in TypeScript. In `codegen.ts`, add the following:

```ts
const config: CodegenConfig = {
    ...
    generates: {
        "./generated/gql/": {
            ...
            config: {
                ...
                scalars: {
                    DateTime: "string",
                }
            },
```

Then restart your frontend container and the ESLint error should go away.

frontend/components/studyDetail/StudyDetailsSidebar.vue
```diff
@@ -1,10 +1,14 @@
 <script setup lang="ts">
 import type { GetStudyQuery } from "~/generated/gql/graphql";
+import { localDateTime } from "../../composables/useLocalisation";
```
XanderVertegaal: Pro tip: use imports with `~` instead of `../../../../../../etc.`, courtesy of your buddy Nuxt:


```suggestion
import { localDateTime } from "~/composables/useLocalisation";
```

backend/research/models.py
```diff
@@ -27,6 +25,9 @@ class Study(models.Model):
...
+        return UserFormSubmission.objects.filter(study=self).last().updated_at
```
XanderVertegaal:
```suggestion
        return UserFormSubmission.objects.filter(study=self).latest('updated_at').updated_at
```

This also gets rid of the type error.

frontend/components/form/FormStepper.vue
```diff
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { useTranslateableAttribute } from "~/composables/useTranslation";
+import { useTranslateableAttribute } from "../../composables/useLocalisation";
```
XanderVertegaal:
```suggestion
import { useTranslateableAttribute } from "~/composables/useLocalisation";
```

frontend/composables/useLocalisation.ts
```diff
@@ -20,3 +20,10 @@ function useTranslateableAttribute<Key extends string>(
 }
 
 export { useTranslateableAttribute };
+
+const timeFormat: Intl.DateTimeFormatOptions = { month: 'long', day: 'numeric', year: 'numeric', hour: 'numeric', minute: 'numeric'}
+function localDateTime(isoString: string) {
+    // localizes the isoString (from for example UTC time) into local datetime and handles translation
```
XanderVertegaal: Good idea to add a comment here, but it is currently just a regular comment while it could be a docstring that we can see when we hover over the function name (as in Python). To do that, have a look at: https://michaelcurrin.github.io/dev-cheatsheets/cheatsheets/javascript/general/jsdoc.html. (Don't bother with the @example, @param etc. Using `/** */` is enough :))

XanderVertegaal (commented):
Two fixes in one PR! Nice :)

I have a few styling-related comments and nitpicks. After addressing those and @EdoStorm96's comment, this is good to go.

frontend/components/studyDetail/StudyDetailsSidebar.vue
RoordinkFrank: This indeed solves it. Must have quite a bit of work to find this out.

backend/form/services/update_submission.py
RoordinkFrank: Fixed.
Line 33 and 43 can be replaced by just triggering the update on line 30, in that case we only have two points where we update so a new variable is for now more lines of code then just calling updated_at directly.


On a sidenote, I feel like submitted at really should be edited at when not yet submitted. But we may need to wait for actual revisions to change that.

EdoStorm96 (approved):
I think this looks good once you solve the merge conflicts!

PR #251, opened 2026-05-12
Chore/reorganize research models

XanderVertegaal:
Should be reviewed after #247.


EdoStorm96 (approved):
I have no problem with this ;)

PR #261, opened 2026-05-26
Feature/study filter annotation keys

EdoStorm96:
So here is my long teased list improvement / Study annotation branch.

To summarize the most important changes:
- Lists are improved. We can now filter studies based on their status and faculty. Adding new filters based on questions in the form will be fairly easy from now on.
- This is possible by annotating the answers to important questions to the study. This can be done by adding an "annotation_key" to a question. Its latest response's answer will then automatically get annotated to the study as `Study.{annotation_key}` eg. `Study.faculty`.
- Study.title will also receive a default value if no answer has been provided yet.
- I've updated our dev data creation and our fixtures to include a faculty and title question, as these are now required for the proper rendering of lists.
- Our steps will now always be attached to their form. We can deduce if they are substeps by checking if they have a parent, but the form will always be attached. This is required and has to be set manually when creating a ste
- Our questions are also always attached to a form. This gets set automatically when calling a question's save() method. This is needed to enforce a database level unique constraint between form and annotation_key.

Lots of changed files as I ran into a lot along the way. The most important files to look at are:
- backend/research/types/StudyType.py
- backend/research/models.py
- backend/form/models/form.py
- backend/form/models/questions.py

The other files flow from these changes. Good luck with the review! I am curious to hear your thoughts!


backend/form/models/form.py
XanderVertegaal, line 1: It does feel good to remove all of these properties and keep the models here nice and clean :)

Don't forget to remove all of these (now) unused imports!

backend/form/models/questions.py
```diff
@@ -1,11 +1,43 @@
...
     step = models.ForeignKey(
         "form.Step", on_delete=models.CASCADE, related_name="questions"
     )
+    # synced from step.form on save
+    form = models.ForeignKey(
+        "form.MRForm",
+        on_delete=models.CASCADE,
+        related_name="questions",
+        null=False,
+        blank=True,
```
XanderVertegaal: Is `blank=True` necessary because `editable=False`? Does Django otherwise complain when you try to save a question without a form in Django Admin?

(Just a question out of curiosity, not a request for change :) )

backend/form/models/questions.py
```diff
@@ -1,11 +1,43 @@
...
+
 
 class BaseQuestion(models.Model):
+
+    annotation_key = models.CharField(
+        blank=True,
+        null=True,
+        help_text=mark_safe(
+            "<strong>Only for important questions! When in doubt, leave this "
+            "blank!</strong> If needed, provide a short, descriptive name in "
```
XanderVertegaal, line 20: Could you add a line that explains what the `annotation_key` is used for/what it does?

backend/form/models/questions.py
```diff
@@ -1,11 +1,43 @@
+import re
```
XanderVertegaal: (Unused import)

backend/form/tests/form_test.py
```diff
@@ -1,8 +1,9 @@
 import pytest
 
 from django.db import IntegrityError, transaction
+from django.core.exceptions import ValidationError, ObjectDoesNotExist
```
XanderVertegaal: (Unused imports)

backend/form/tests/form_test.py
```diff
@@ -13,80 +14,85 @@ def test_database_constraint_prevents_invalid_steps(form: MRForm):
...
+        # Creating a second question with the same annotation_key should fail
+        with transaction.atomic():
+            with pytest.raises(Exception):  # IntegrityError from DB constraint
+                TextQuestion.objects.create(
+                    text="Question 2",
+                    step=step,
+                    annotation_key="first_question",
+                )
+
+    def test_annotation_key_allowed_in_different_forms(self):
```
XanderVertegaal, line 76: Nice! 👍

backend/form/admin.py
```diff
@@ -432,6 +437,7 @@ class FileUploadQuestionAdmin(admin.ModelAdmin):
                 "fields": (
                     "text_nl",
                     "text_en",
+                    "annotation_key",
```
XanderVertegaal, line 440: This makes me think we should probably look into deduping the fields in all of these `...QuestionAdmin` classes at some point.

backend/form/queries.py
```diff
@@ -39,3 +40,37 @@ def resolve_form(
...
+        annotation_key=String(),
+        description="Retrieves a SelectQuestion by id or annotation_key, including its options.",
+    )
+
+    @staticmethod
+    def resolve_select_question(
+        root,
+        info: ResolveInfo,
+        id: str = None,
+        annotation_key: str = None,
```
XanderVertegaal:
```suggestion
    @staticmethod
    def resolve_select_question(
        root,
        info: ResolveInfo,
        id: str | None = None,
        annotation_key: str | None = None,
```

backend/form/queries.py
```diff
@@ -39,3 +40,37 @@ def resolve_form(
...
+            return None
+
+        if not id and not annotation_key:
+            return None
+
+        try:
+            if id:
+                question = SelectQuestion.objects.get(id=id)
+            else:
+                question = SelectQuestion.objects.get(annotation_key=annotation_key)
```
XanderVertegaal: What if there are two forms with SelectQuestions with the same `annotation_key`? That would give a `SelectQuestion.MultipleObjectsReturned`.

backend/form/queries.py
```diff
@@ -39,3 +40,37 @@ def resolve_form(
...
+
+            return SelectQuestionType(question=question)
```
XanderVertegaal, line 80: Why is `AttributeError` caught here? When would you expect it?

backend/form/queries.py
```diff
@@ -6,7 +6,8 @@
 from form.services.form_evaluator import FormEvaluator
 from form.services.user_form_resolver import UserFormResolver
 from form.types.UserFormType import UserFormType
-from form.models import UserFormSubmission
+from form.types.QuestionType import SelectQuestionType
+from form.models import UserFormSubmission, BaseQuestion, SelectQuestion
```
XanderVertegaal: (Unused import)

backend/form/migrations/0001_initial.py
```diff
@@ -221,8 +221,8 @@ class Migration(migrations.Migration):
                 (
                     "form",
                     models.ForeignKey(
-                        blank=True,
-                        null=True,
+                        blank=False,
```
XanderVertegaal: This messed up my migration history :') Next time I prefer a new migration (that we may squash later) over time-travelling/rewriting history.

backend/form/fixtures/vwr.json
```diff
@@ -63,10 +71,12 @@
         "model": "form.basequestion",
         "pk": 5,
         "fields": {
+            "annotation_key": "faculty",
```
XanderVertegaal, line 74: I like that you use the new `annotation_key` sparingly in the fixtures 👍.

backend/form/fixtures/vwr.json
```diff
@@ -764,23 +862,23 @@
     },
     {
         "model": "form.selectquestion",
-        "pk": 34,
+        "pk": 39,
```
XanderVertegaal, line 865: Why are these questions switched? Not a problem as far as I'm concerned; just curious.

backend/main/management/commands/create_dev_data.py
```diff
@@ -224,6 +228,10 @@ def _create_select_question(form: Step, index: int) -> None:
                 multiple=self.faker.pybool(),
             )
 
+            if SELECT_ANNOTATION_KEYS:
+                select_question.annotation_key = SELECT_ANNOTATION_KEYS.pop()
```
XanderVertegaal, line 232: Clever way to make sure each key is used only once! 👍

backend/main/management/commands/create_dev_data.py
```diff
@@ -259,31 +270,8 @@ def _create_file_upload_question(form: Step, index: int) -> None:
...
-                all_steps.extend(_get_all_substeps(step))
-
-            return all_steps
-
-        all_steps = _get_all_steps(form)
-
         for step in tqdm(
-            all_steps, desc="Generating questions...", disable=options["silent"]
+            form.steps.all(), desc="Generating questions...", disable=options["silent"]
         ):
```
XanderVertegaal, line 275: Truly nitpicking, but you could add a `# type: ignore` at the end of this line and 384, so the type checker remains 'happy' with this file.

XanderVertegaal: To be able to simply do `form.steps` without all of the gymnastics above is great, btw <3

backend/api/graphql/schema.py
```diff
@@ -47,5 +48,6 @@ class Mutation(ObjectType):
         SelectQuestionType,
         FileUploadQuestionType,
         NumberQuestionType,
+        GQLSubmissionStatus,
```
XanderVertegaal: Do you need to add GQLSubmissionStatus here? You've added this as a field to the StudyType, so Graphene should pick up on it. The Study detail page seems to work well without this line.

backend/research/models/reviews.py
```diff
@@ -5,10 +5,12 @@
 
 
 class SubmissionStatus(models.TextChoices):
-    DRAFT = "DRA"
-    SUBMITTED = "SUB"
-    APPROVED = "APP"
-    REJECTED = "REJ"
+    # NOTE: These names need to be identical in order for codegen
```
XanderVertegaal, line 8: Sad and weird but true. 🤨

backend/research/models/study.py
```diff
@@ -17,6 +19,99 @@ def _viewable_objects(self, user: User):
...
+
+        Assumes title is already annotated by with_answers_by_annotation_key_annotation.
+        """
+        if queryset is None:
+            queryset = super().get_queryset()
+
+        # create a default title in case there is not yet an answer to the title question
+        default_title_str = Concat(
+            Value("Study created on "),
+            Cast(Extract("created_at", "year"), CharField()),
```
XanderVertegaal, line 40: Can we do DD-MM-YYYY instead (incl. padding with 0, as you do with the minutes below)

(Kudos for this database function magic, btw!)

backend/research/models/study.py
```diff
@@ -1,11 +1,13 @@
 from form.models import UserFormSubmission
 from main.models import User
-from form.models import MRForm, QuestionResponse, UserFormSubmission
+from form.models import MRForm, BaseQuestion, QuestionResponse, UserFormSubmission
```
XanderVertegaal: (Unused imports)

backend/research/models/study.py
```diff
@@ -17,6 +19,99 @@ def _viewable_objects(self, user: User):
...
 
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        # Annotate important fields (aka question with string_id)
+        queryset = self.with_answers_by_annotation_key(queryset)
+        # Annotate a default title, if no title
+        queryset = self.with_default_title_annotation(queryset)
+        return queryset
+
+    def with_default_title_annotation(self, queryset=None):
```
XanderVertegaal: Is there a case in which `queryset` would be `None`? (The same goes for the function below btw.)

backend/research/models/study.py
```diff
@@ -17,6 +19,99 @@ def _viewable_objects(self, user: User):
...
+
+        This is used as a default when getting a queryset in get_queryset.
+
+        Args:
+            queryset: Optional queryset to annotate (defaults to all studies)
+
+        Returns:
+            Annotated queryset with dynamic fields named after question annotation_keys
+        """
+        if queryset is None:
```
XanderVertegaal: I'm not sure this is needed.

backend/research/models/study.py
```diff
@@ -17,6 +19,99 @@ def _viewable_objects(self, user: User):
...
+            .filter(annotation_key__isnull=False)
+            .exclude(annotation_key="")
+            .distinct()
+        )
+
+        # Build annotations dictionary
+        annotations = {}
+        for question in questions:
+
+            answer = (
```
XanderVertegaal, line 102:
```suggestion
        for question in questions:
            answer = (
```

backend/research/tests.py
```diff
@@ -201,3 +202,164 @@ def test_concurrent_save_generates_unique_references(self, normal_user: User):
...
+        response2 = QuestionResponse.objects.create(
+            question=q2,
+            answer={"value": "Berlin"},
+        )
+        response2.submissions.add(submission)
+
+        # Retrieve study from queryset with annotations
+        study_with_annotations = Study.objects.get(pk=study.pk)
+
+        # Check that annotations are present
```
XanderVertegaal, line 256: Nice!

frontend/pages/studies/index.vue
```diff
@@ -79,35 +85,53 @@ const orderingOptions = computed(() => {
...
     };
-    return [userFilter];
+    return [facultyFilter, statusFilter];
 });
 </script>
 
 <template>
-    <h4>Note: This page is a proof-of-concept for future lists.</h4>
     <SharedGraphQLList
+        v-if="!facultyLoading"
```
XanderVertegaal, line 139: For extra brownie points you could add a `v-else` with a loading spinner or something fancy.

Absolutely not something I insist on.

XanderVertegaal (approved):
Definitely an improvement, really well done! <3 

I have a few vanity points/nitpicks but nothing that should stop you from merging.

backend/form/queries.py
EdoStorm96: Yes this is a bit of a tricky one ... I think my preferred solution would be to make SelectOption.question a M2M field. So then we can reuse selectoptions across selectquestions. Then if we get two questions with the same annotation_key here, we can just merge their select options (this would then become a selectoption query instead of a selectquestion query, with the question's annotation_key as the argument). But it really depends on how we are planning to do the form versioning. I still am not 100% on that.

I could do this in this PR ... But maybe it'd be better to fix this later. For now I could just return that latest selectquestion if I retrieve multiples? Or add the form to the query. What do you think?

frontend/pages/studies/index.vue
EdoStorm96: Gonna skip this one, as it loads pretty fast

backend/form/fixtures/vwr.json
EdoStorm96: I guess Django dumpdata is not entirely consistent ... I am looking into this, but not in this PR.

PR #263, opened 2026-06-03
feat: email type added.

### LINKED ISSUE #133: E-mail question type
So the frontend knows to validate it as such

RoordinkFrank:
fixes #133 

Adds is_email field and frontend validation.
The frontend email validation is not the same as the (future) django backend validation. 

backend validation will be added in #254

I have tried to convert the Django regex to Typescript but to no result as EmailValidator (The validator that Django Email Field uses) uses regex that is not recognized in JavaScript (A few key differences are \Z and $ matches, octal to hex numbering that are different.). Some of it an be translated but some of it simply doesn't work and would have to be removed from the regex.

At that point if they cann´t be the same regex than it's better to simply call a JavaScript validator.

Django's backend regex which would be 
`@staticmethod
    def email_regex() -> str:
        validator = EmailValidator()
        full_pattern_string = validator.user_regex.pattern + "@" + validator.domain_regex.pattern
        return full_pattern_string`

or call the validator with something like this

```
    def is_valid_email(self):
        temp_email = "hello@roordink.nl"
        validator = EmailValidator()
        validator(temp_email)
```
---------------------------

vwr.json will be regenerated and reformatted once rich-text-formatting-in-questionstep-descriptions has been merged in develop. It only adds is_email = false to all textquestions.


frontend/locales/nl.json
```diff
@@ -1,133 +1,134 @@
...
+        " number": "Ref. nummer"
+    },
+    "An error occurred while saving the form. Please try again.": "Er is een fout opgetreden bij het opslaan van het formulier. Probeer het opnieuw.",
+    "Registration submitted successfully.": "Registratie succesvol ingediend.",
+    "Your form contains errors. Please review and resubmit.": "Het formulier bevat fouten. Bekijk deze en probeer het opnieuw in te dienen.",
+    "Confirm new registration": "Bevestig nieuwe registratie",
+    "Are you sure you want to register a new study?": "Weet je zeker dat je een nieuwe studie wilt registreren?",
+    "Delete": "Verwijderen",
+    "No actions available": "Geen acties beschikbaar",
+    "This is not a valid email": "Dit is geen valide email"
```
```suggestion
    "This is not a valid email": "Dit is geen geldig e-mailadres."
```

backend/form/models/questions.py
```diff
@@ -1,4 +1,9 @@
+from re import Pattern
```
XanderVertegaal: Unused imports.

backend/form/types/QuestionType.py
```diff
@@ -106,6 +106,7 @@ class Meta:
     placeholder_nl = String()
     placeholder_en = String()
     lines = Int(required=True)
+    is_email = Boolean(required=False)
```
XanderVertegaal: I think this can/should be `required=True`. For every `TextQuestion` object, this field has a value.

frontend/composables/useProcessForm.ts
```diff
@@ -323,6 +323,17 @@ function addValidationRule(
                 );
             }
     }
+    switch (question.__typename) {
+        case "TextQuestionType":
+            if (question.isEmail) {
+                rules.isEmail = helpers.withMessage(
+                    t("This is not a valid email"),
+                    (value: string) => validator.isEmail(value),
```
XanderVertegaal: Vuelidate comes with its own [email validator](https://vuelidate-next.netlify.app/validators.html#email). Could you try using that instead? If it doesn't work, we can always go for an additional dependency, but if we don't have to, I'd prefer that :)

backend/form/fixtures/vwr.json
XanderVertegaal, line 1: Please keep the indentation the same. Also, some questions seem to have switched order. Is that intentional?

XanderVertegaal (changes requested):
Good work! Nice to see that the required code changes for this feature are relatively minimal.

Have a look at my comments, especially at the one about using Vuelidate's built-in email validator. In addition:
- `<input>` elements can have their type attribute set to "email" so the browser can also do some validation. I'm not sure how useful it is, or combines with our own validation system, but you might want to try it: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/email
- Update the `create_dev_data` script so some text questions are created with `is_email` set to True.

backend/form/fixtures/vwr.json
RoordinkFrank: Same problem as with #175, The current vwr.json seems to be wrong. A separate PR by Edo or Me will be made to address this.

The problem is that _order does not matter in the fixture. Just which object comes first top to down in the file itself. And those two are not in sync right now. Those _order changes you see are just corrections for what is already reality.

You should be able to validate this by loading the fixtures and without changes recreating the fixtures. Right now they should not be the same.

frontend/composables/useProcessForm.ts
RoordinkFrank: Good find. I did not expect Vuelidate to have one.

backend/form/types/QuestionType.py
RoordinkFrank: You are right. I think I confused this with the default setting here when I wrote this. But that of course is in model.

backend/form/fixtures/vwr.json
RoordinkFrank: Temp putting this on hold until PR is available.

RoordinkFrank:
Everything works well and is much simpler now. Thanks for the hint that vuelidate already had an email.

--------

image


You mean like in NumberQuestion.vue here.
```
<input
            :id="`${question.questionId}-${question.repeatIndex}`"
            v-model="modelValue"
            type="number"
            class="form-control"
            :class="{ 'is-invalid': isInvalid }"
            :min="question.positiveOnly ? 0 : undefined"
        />
```

I can see it but in the end this is not what triggers the error message. Only useProcessForm does.

backend/form/fixtures/vwr.json
```diff
@@ -2411,132 +2411,145 @@
     },
     {
         "model": "form.textquestion",
-        "pk": 26,
```
RoordinkFrank, line 2414: Pk's do not get changed but the order in the file does.
Command used is `sudo docker compose exec django-dev python manage.py create_form_fixture`

frontend/composables/useProcessForm.ts
```diff
@@ -323,6 +325,15 @@ function addValidationRule(
                 );
             }
     }
+    switch (question.__typename) {
+        case "TextQuestionType":
+            if (question.isEmail) {
+                rules.isEmail = helpers.withMessage(
+                    t("This is not a valid email"),
```
RoordinkFrank, line 388: Image

Remember translations with question type specific rules do not get triggered until a full f5. I do not know if there is already an issue for this.

EdoStorm96 (changes requested):
Mmm ... Just at a quick glance, I think it would be better to create a new subclass of BaseQuestion named EmailQuestion. In my mind, that would not need to have any specific attributes, so it could just be an empty subclass of BaseQuestion. But that would be cleaner IMO then to repurpose TextQuestion for EmailQuestions, because it saves us having an is_email attribute on all TextQuestions.

RoordinkFrank:
> Mmm ... Just at a quick glance, I think it would be better to create a new subclass of BaseQuestion named EmailQuestion. In my mind, that would not need to have any specific attributes, so it could just be an empty subclass of BaseQuestion. But that would be cleaner IMO then to repurpose TextQuestion for EmailQuestions, because it saves us having an is_email attribute on all TextQuestions.

That was my first hunch as well but after a discussion with Xander I changed it to TextQuestion.

XanderVertegaal:
Yes, the idea to have the TextQuestion do double duty is on me. I do not feel adamant about it, so @EdoStorm96 if you prefer to have a separate model, I wouldn't be against a refactor :)

RoordinkFrank:
I will refactor this back to email question.

backend/form/fixtures/vwr.json
EdoStorm96, line 1: Did you generate this with our create_form_fixture script?

frontend/package-lock.json
EdoStorm96, line 1: Could you remove these changes from this branch? Best to keep our dependency changes separate from our feature branches.

EdoStorm96 (approved):
I think it looks all good! Nice work. Once you fix the merge issues, you should be ready to merge this (once all checks are passed too ;p)

backend/form/fixtures/vwr.json
EdoStorm96: Sorry, I see you did.

PR #265, opened 2026-06-05
Feature/faq page using simple notes app

RoordinkFrank:
fixes #241

to test

create_dev_data creates random slugs. You can visit those slugs with /notes/slugname.

To test in graphQL

```
query getNote($slug: ID!) {
  note(slug: $slug) {
    titleNl
    titleEn
    contentEn
    contentNl
    slug
  }
}

```

```
{
  "slug": "slugname"
}

```

----------------------------------

I originally misunderstood what the use case was for this notes app and I tried to implement the notes so they can be used in for example the frequentQuestions.vue component.

My idea was that the title would be the accordion button and the content would be the accordion content.

image

It would need this in the queries.py file to function. No tag in the parameters means you get all notes.

```
    notes = List(
        NonNull(
            NoteType,
        ),
        tag=String(required=False),
    )
```

```

    @staticmethod
    def resolve_notes(root, info: ResolveInfo, tag: str = "") -> QuerySet[Note]:
        queryset = NoteType.get_queryset(Note.objects, info).all()
        if tag:
            queryset = queryset.filter(tag=tag)
        return queryset
```

I still think that this is a good idea to create but that will be a new issue. Frequent questions is basically the same as notes except in a slightly different interface made for the frontpage. This of course also depends what the plans are for #244

--------------------

There is also heavy overlap with Step Info Texts, perhaps they could be combined in the future? The tag could be reused to reference the step.


backend/notes/admin.py
```diff
@@ -0,0 +1,41 @@
...
+            'cdh.core/js/jquery-3.6.1.min.js',
+            'cdh.core/js/tinymce/tinymce.min.js',
+            'cdh.core/js/tinymce/tinymce-jquery.min.js',
+            'cdh.core/js/tinymce/shim.js',
+        )
+    formfield_overrides = {
+        models.TextField: {"widget": TinyMCEWidget},
+    }
+
+    @admin.display(description="Note URL")
```
RoordinkFrank: Note_link is longer in use as it relies on url and view files that are not used with graphQL. Still it looks neat and perhaps we can link this to a new frontpage version that right now does not exist yet.

backend/notes/models.py
EdoStorm96, line 1: I like this! Nice and minimal :)

backend/notes/models.py
```diff
@@ -0,0 +1,29 @@
...
+class Note(models.Model):
+    """
+    A very barebones model, which can hold html-formatted text, created in the
+    admin interface using TinyMCE
+    """
+
+    title = models.CharField("Note title", max_length=30)
+    content = models.TextField("Note content",
+                               help_text="For links to other pages within the "
+                               "MR/grant-tool, format url like '/notes/1/'. For links, "
```
EdoStorm96: I would just say "For links to other pages within MyResearch"

Because the grant-tool is an external app for the foreseeable future, this is a bit confusing.

frontend/pages/notes/[slug].vue
```diff
@@ -0,0 +1,36 @@
...
+</script>
+
+<template>
+    <div v-if="note" class="uu-content">
+        <div class="uu-hero">
+            <h1>{{ useTranslateableAttribute(note, "title") }}</h1>
+        </div>
+        <div
+            class="uu-container"
+            v-html="useTranslateableAttribute(note, 'content')"
```
Comment (EdoStorm96) [inline frontend/pages/notes/[slug].vue]: Could you add a child div to the uu-container with `class="col-9"` and have the v-html here? For two reasons: The uu-container is a flex-row by default, which means it will try to add things horizontally. I you, in tiny mce, create a note with a short head and a short paragraph, these will get placed next to eachother. Adding the col class makes a column, and I personally think it looks nice to not have the text be the full width of the page. We then go from this: 

Image
To this:

Image

EdoStorm96 (approved):
This is perfect in my opinion! Nice and simple. I would just like you to add the CSS i suggest and then this can be merged as far as I'm concerned. Nice work!

PR #267, opened 2026-06-09
fix: ensure NUXT_PUBLIC_SAML_URL is available in prod

EdoStorm96:
Tiny fix to get the login buttons working on prod.


miggol (approved):
Makes perfect sense.

If you feel like it you could add the updated `/frontend/schema.json` to this PR. Approved either way.

EdoStorm96:
@miggol I've added the schema and also ensured that codegen runs when building prod, based on the schema. This should fix the error we encountered when deploying yesterday.

.github/workflows/build-and-test-django.yml
```diff
@@ -21,4 +21,4 @@ jobs:
       - name: Check for missing migrations
         run: docker compose run django-prod python manage.py makemigrations --dry-run --check
       - name: Run Django tests
-        run: docker compose run django-prod pytest
+        run: docker compose run django-prod pytest
```
miggol: Missing newline at end of file >:(

.github/workflows/check-schema.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: docker compose --profile dev build django-dev
+      - name: Generate GraphQL schema
+        run: docker compose --profile dev run django-dev python manage.py graphql_schema --out /usr/src/app/backend/schema_generated.json
+      - name: Check backend/schema.json is up to date
+        run: diff backend/schema.json backend/schema_generated.json
+      - name: Check frontend/schema.json is up to date
+        run: diff frontend/schema.json backend/schema_generated.json
+      - name: Cleanup
+        if: always()
+        run: rm -f backend/schema_generated.json
```
miggol: AFAIK this isn't necessary. Github runners are ephemeral and are destroyed after use, including anything on disk and built images (that you didn't explicitly put in a registry or whatever).

.github/workflows/check-schema.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: docker/setup-compose-action@v1
+      - name: Copy test environment and secrets
+        run: |
+          cp .env.example .env
+          cp certificates/private.key secrets/
+          cp certificates/public.cert secrets/
+      - name: Build django-dev container
+        run: docker compose --profile dev build django-dev
+      - name: Generate GraphQL schema
+        run: docker compose --profile dev run django-dev python manage.py graphql_schema --out /usr/src/app/backend/schema_generated.json
```
miggol: Building the docker container is a costly process that takes up a lot of minutes. I think we should put as much as possible in the other workflow that already builds the container.

Although prod doesn't mount a volume, there are other ways to get output from it. For example, you can use `manage.py graphql_schema --out=-` to print the schema to stdout. This should be the exact same text that is otherwise written to `schema.json`.

To use it, combine it with `docker-compose run --quiet` so that docker outputs nothing other than the command output, and then diff that with the schema in the repo.

Or write a little shell script or management command that checks the schema inside the container, it's simple and it works. The `diff` command is available by default inside `django-prod` so you could also get it to work with a multiline `docker-compose run` command.

miggol (changes requested):
Quick review mostly looking at the workflow. I would prefer we did this check in the other workflow.

The codegen and nuxt fix look good though!

.github/workflows/build-and-test-django.yml
```diff
@@ -1,24 +1,30 @@
...
+                  cp certificates/public.cert secrets/
+            - name: Build django-prod container
+              run: docker compose build django-prod
+            - name: Check for missing migrations
+              run: docker compose run django-prod python manage.py makemigrations --dry-run --check
+            - name: Check schemas are up to date
+              shell: bash
+              run: |
+                  generated=$(docker compose run -T --quiet django-prod python manage.py graphql_schema --out=-)
+                  diff <(printf '%s' "$generated") backend/schema.json
```
miggol, line 27: This turned out way neater than I imagined. Kudos!

backend/main/management/commands/create_dev_data.py
```diff
@@ -413,14 +413,15 @@ def _create_user_form_submission(
...
+            range(10),
+            desc="Generating notes...",
+            disable=options["silent"],
         ):
             note = Note.objects.create(
                 title_nl=self.faker_nl.sentence(nb_words=1),
                 title_en=self.faker_en.sentence(nb_words=1),
                 content_nl=self.faker_nl.paragraph(),
                 content_en=self.faker_en.paragraph(),
+                slug=self.faker.unique.slug(),
```
miggol, line 425: Thanks, this was bothering me. Fixes #290

miggol (approved):
Accidentally pushed straight to your branch. I moved the build context for vue-prod one directory up which works fine. I chose to remove the frontend schema for convenience, as the backend version is the default output path.

Give it a once-over in case I messed anything up but otherwise this should be good to merge!

PR #271, opened 2026-06-11
Feature/file upload fields

XanderVertegaal:
Closes #242 

Implements the FileUploadQuestion using DSC's `FileField`, which is hosted by a new model `MRDocument`. Please test this carefully; there are quite some moving parts that work together here.

image

- The GraphQL spec does not account for file uploads, so I have added two new REST views to upload and download files.

- As documented in the README, the value of QuestionResponse.answer to a FileUploadQuestion has the following shape:
  ```json
  FileUploadQuestion: {
      "value": "21cbf05a-987d-443b-b091-7c29e40dee92", 
      "name": "important.pdf", 
      "size": "500000"
  }
  ```
  This is all the information we need to display the title and file size to the user whenever they revisit the page without downloading the file right away. If the user wants to see the file, the UUID in `answer.value` is used to retrieve the actual file.

- A frontend validation rule has been added that prevents the user from submitting if the file size exceeds `FileUploadQuestion.size_limit`.

  image

- Lastly, `useConditionalQuestionsWatcher` is renamed to `useWatchQuestions` and FileUploadQuestions are now included in the list of watched questions that immediately trigger form resubmission. We need this to make sure the UserFormSubmission is updated every time a file is uploaded or removed. 

Bonus:
- The GraphQL graphical interface was always enabled. It is now generally disabled for PROD and enabled for DEV. I was dealing with the CSRF token anyway, so it made sense to make this small fix here.


backend/form/models.py
XanderVertegaal, line 1: I think this file remained from some earlier PR. It was empty on my system.

backend/main/management/commands/create_dev_dataset_test.py
```diff
@@ -23,6 +24,7 @@
     Review,
     ReviewRound,
     YearCounter,
```
XanderVertegaal, line 26: I can only properly mock MRDocument if I create a dummy file somewhere. It works, but it adds file handling overhead and slows down the create_dev_data script, so I decided to leave this until we actually need it.

backend/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM python:3.11-alpine
 # Add git to install some requirements
 # Add curl to use for container healthcheck
 # Add xmlsec for saml
-RUN apk add --no-cache git curl xmlsec
+RUN apk add --no-cache git curl xmlsec libmagic
```
EdoStorm96, line 6: This was needed to get it working?

backend/form/migrations/0014_alter_fileuploadquestion_size_limit_mrdocument.py
EdoStorm96, line 1: Might have to regenerate migrations before merge ...

backend/form/views.py
```diff
@@ -1,3 +1,65 @@
...
+    Access is granted to the submitting user, privacy officers and FETC members.
+    """
+
+    def get(self, request, **kwargs):
+        if self._file_wrapper is None:
+            return HttpResponseNotFound()
+
+        uuid = self.kwargs.get(self.uuid_path_parameter)
+        try:
+            document = MRDocument.objects.get(file__uuid=uuid)
```
EdoStorm96, line 48: You could implement the accessible objects way of doing permissions for MRDocuments. But in the interest of time, I also am fine with letting this slide. This does seem to work well.

backend/form/types/QuestionType.py
EdoStorm96, line 214: Maybe place a comment that this is in bytes. (also could you change the fixture to have a size limit of like 10.000.000 aka 10mb or whatever seems suitable? Now it's 500 bytes.

EdoStorm96 (approved):
I think this all looks good and it is surprisingly more simple than I thought. Very nice IMO. I've tested this and it works really well. I have some minor comments, but all I'd really like you to do is change the size limit in the fixture. Nice work!

backend/Dockerfile
XanderVertegaal: Yes! Good that you ask, but it's really necessary :P

backend/form/types/QuestionType.py
XanderVertegaal: The dev server is no place to upload cat videos to. 500 bytes is all you get >:(

(No, of course I'll change this!)

backend/main/management/commands/create_dev_dataset_test.py
miggol: I recommend the official standardized PortalDev [test PDF](https://github.com/DH-IT-Portal-Development/ethics/blob/develop/media/test_pdf.pdf)

backend/form/views.py
XanderVertegaal: Good idea! Let's pick that up in a future refactor.

miggol (approved):
Looks good!

PR #274, opened 2026-06-16
feature: backend validation added

RoordinkFrank:
fixes #254

Wants to merge into feature/e-mail-question-type.

Adds:
backend valdiation of positive number
backend validation of email type
backend validation of future only dates
frontend validation of future only dates

Does not add:
file size_limit validation in backend. As this is not an optional boolean it can be called in multiple ways validate/clean/validation_rule. I have opted out of implementing this as #271 is still ongoing.

-----------------------
Positive number can be tested with:
Change the number 0 to 1 in the question model if statement. You should be able to find this error in the console.
image

--------------------

I took the liberty to also add frontend validation for future date on top of backend validation. This is outside the scope of the issue but the form would will not work for test users without this. I could have create a separate issue for this but we have only got two weeks left. 

To test the backend temporary remove the frontend validation. Also make sure to mark a date as future_only as that is not yet activated in vwr.json. Email testing goes similar.

-------------------

As a last note frontend and backend email validation can theoretically not be in sync though it will be rare in practise. The only solution I can think of is to print out the backend errors on the frontend form. Or just as a notification error.


backend/form/models/questions.py
```diff
@@ -1,41 +1,12 @@
...
-from django.core.exceptions import ValidationError
 from django.db import models
-
-from django.utils.safestring import mark_safe
-
-snake_case_validator = RegexValidator(
-    regex=r"^[a-z]+(_[a-z]+)*$",
-    message="Only snake_case is allowed (e.g. 'recording_details'). Read help text carefully!",
-)
+from django.core.exceptions import ValidationError
```
RoordinkFrank, line 3: Was missing in clean method.

backend/form/models/questions.py
RoordinkFrank, line 68: There is this clean() method that as far as I can see is not called right now. I am quite certain about that as the ValidationError was not yet imported so it should not have worked.

That said I am uncertain if clean() or validate() should be used. Clean() is usually used in Django forms (even though this is in models) and what we have is a kind of form even though we aren´t really cleaning anything the method still has a bit of Django support.
We will need to make a choice here what to use, personally I think validate() is fine as itś more specific on what it does right now.

----------

Edit: I just saw that a questions.save() method has been added from another branch. I guess it does get called now.

backend/form/services/update_submission.py
```diff
@@ -22,6 +22,8 @@ def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSub
 
     for response in user_form_input.get("responses", []):  # type: ignore
         response_id = response["id"] if "id" in response else None
+        # Immediately cut off responses that are not valid.
+        validate_response(response)
```
RoordinkFrank, line 48: The only thing I am uncertain about is if the responses from earlier in the for loop are actually saved or not. Though I suppose it doesn´t really matter as they are correct responses.

RoordinkFrank:
Set to draft again. I just discovered we have a future only date option, so that one will need implementation as well.

edit: frontend and backed for future only dates added.

backend/form/services/update_submission.py
```diff
@@ -52,3 +61,29 @@ def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSub
...
+    question = get_question(question_id)
+    # validate will throw an error in case of wrong input.
+    question.validate(value)
+
+
+def get_question(
+    question_id: int,
+) -> BaseQuestion | NumberQuestion | TextQuestion | DateQuestion:
+    """Searches child models of BaseQuestion and returns the appropriate question"""
+    # This answer will always be the same in the current form version.
```
RoordinkFrank, line 99: This would also be easier if question type gets returned to the backend (though that will mean more data) but for now this does work.

backend/form/models/questions.py
```diff
@@ -127,14 +131,33 @@ class TextQuestion(BaseQuestion):
...
+
 
 class NumberQuestion(BaseQuestion):
     positive_only = models.BooleanField(default=False)
 
+    def validate(self, answer: str):
+        super().validate(answer)
+        value = int(answer)
+        if self.positive_only and value < 0:
+            raise ValueError("value in must be positive")
```
RoordinkFrank, line 149: These errors are mostly for us only so translation isn't needed.

backend/form/models/questions.py
```diff
@@ -79,6 +80,9 @@ def save(self, *args, **kwargs):
         self.full_clean()
         super().save(*args, **kwargs)
 
+    def validate(self, answer: str):
```
EdoStorm96, line 83: Maybe you can already validate required here?

backend/form/models/questions.py
```diff
@@ -127,14 +131,33 @@ class TextQuestion(BaseQuestion):
     lines = models.PositiveIntegerField(default=1)
     is_email = models.BooleanField(default=False)
 
+    def validate(self, answer: str):
+        super().validate(answer)
+        validator = EmailValidator(message="value must be a valid email address")
+        # The email validator in the frontend is slightly different.
```
EdoStorm96, line 137: What is the difference? Might be nice to also include in the comment.

backend/form/models/questions.py
```diff
@@ -127,14 +131,33 @@ class TextQuestion(BaseQuestion):
     lines = models.PositiveIntegerField(default=1)
     is_email = models.BooleanField(default=False)
 
+    def validate(self, answer: str):
+        super().validate(answer)
+        validator = EmailValidator(message="value must be a valid email address")
```
EdoStorm96, line 136: You should probably instantiate the validator after the if statement.

backend/form/services/update_submission.py
```diff
@@ -52,3 +61,29 @@ def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSub
             new_response.submissions.add(current_submission)
 
     return current_submission
+
+
+def validate_response(response):
+    """Backend validation incase malicious responses. Under normal circumstances all validations are already checked in the frontend"""
+    question_id = response["question_id"]
+    answer = response["answer"]
```
EdoStorm96, line 88: I'd prefer `value = response["answer"]["value"]`

backend/form/services/update_submission.py
```diff
@@ -52,3 +61,29 @@ def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSub
...
+    value = answer["value"]
+    question = get_question(question_id)
```
EdoStorm96, line 90: And I'd prefer `question = get_question(response["question_id")`

EdoStorm96: Also, get_question is not necessary, it basically does the same as BaseQuestion.get_subclass(). You can just do: `question = BaseQuestion.objects.get(response["question_id"]).get_subclass()`

backend/form/services/update_submission.py
```diff
@@ -52,3 +61,29 @@ def update_submission(user: User, user_form_input: UserFormInput) -> UserFormSub
...
+    # validate will throw an error in case of wrong input.
+    question.validate(value)
+
+
+def get_question(
```
EdoStorm96, line 95: This can be removed as it mostly does the same thing as BaseQuestion.get_subclass()

frontend/composables/useProcessForm.ts
```diff
@@ -334,6 +334,15 @@ function addValidationRule(
                 );
             }
     }
+    switch (question.__typename) {
```
EdoStorm96, line 393: This also goes for the question.is_email rule, but this is not how you use switch-case. Have a look at: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch). You only need one switch statement, and then multiple case statements. So have switch (question.__typename) and then have multiple cases for different names.

backend/form/models/questions.py
EdoStorm96: I also think validate is fine for our special case :) Always a bit confusing when top use which validation method ...

backend/form/models/questions.py
EdoStorm96: Be sure, if you do this, to do this in the same way for True False questions as Xander did it in #276

EdoStorm96 (commented):
I have some comments, but I think the general design is not bad actually. I don't really see a problem here and it is nice that you added it!

I would appreciate if @miggol or @XanderVertegaal can have a second look at this, because maybe I am missing something. But I think this is a nice feature!

backend/form/services/update_submission.py
RoordinkFrank: Did not know that was possible. Certainly looks cleaner.

backend/form/services/update_submission.py
RoordinkFrank: I will take a look into this. If it works it would certainly be way more convenient.

frontend/composables/useProcessForm.ts
RoordinkFrank: You are right. I was absolutely not paying attention here just copy paste from above and mostly looking at the helpers.withMessage. Very stupid mistake.

PR #275, opened 2026-06-16
Feature/custom dumpdata

EdoStorm96:
A custom command to dump form fixtures, to that the _order of questions, steps and selectoptions is consistent. Read my docstring in create_form_fixture for more info ;p

We should use this to update fixtures from now on, because otherwise Django will mess with our _order.


XanderVertegaal (approved):
Great, this works well as far as I can tell. If you can avoid `dumpdata_temp.json` being created in the container (and maybe add a smoke test as Copilot suggests), this is good to go :)

PR #276, opened 2026-06-16
Feature/validate true false questions

XanderVertegaal:
TrueFalseQuestions always have a value, so Vuelidate's built-in `required` validator always passes and questions like "I've understood the above" never block submission.

A special addition is made to `useProcessForm()`: required TrueFalseQuestions should have an answer with value `True` to be admitted.


EdoStorm96 (approved):
Yes, this makes sense. I have not tested this, but presuming you have, I approve. I would appreciate a bit more explanation for requiredFn though.

PR #282, opened 2026-06-18
fix/dynamic errormessages in the frontend

RoordinkFrank:
fixes #264 

Wants to merge into feature/validate-submitted-answers-in-the-backend. There are a few more message that need this fix in that branch.


XanderVertegaal (approved):
I haven't tested this functionally, but the changes look good, so if you tested this and it works, feel free to merge!

PR #289, opened 2026-06-23
Feature/polish homepage

RoordinkFrank:
fixes #244
The homepage will not be done after this. This is to make it acceptable to the test users.


frontend/components/frontpage/FrequentQuestions.vue
```diff
@@ -1,86 +0,0 @@
-<script lang="ts" setup>
-import { mockQuestions } from "~/shared/mockData";
```
RoordinkFrank, line 2: I do wish to one day reuse this but for now the FrequentQuestions has been moved to a FAQ note. Leon has asked for the FAQ to be clickable. As in you click on the questions and the answer appears. Something this does. But that's something for after 1 july. So for now goodbye.

backend/notes/fixtures/initial.json
```diff
@@ -0,0 +1,15 @@
+[
+{
+    "model": "notes.note",
+    "pk": 11,
+    "fields": {
+        "title": "Frequent Questions",
+        "title_en": "Frequent Questions",
+        "title_nl": "Veelgestelde Vragen",
+        "content": "<p>English FAQ nog yet implemented</p>",
+        "content_en": "<p>English FAQ nog yet implemented</p>",
```
RoordinkFrank: You are better off reviewing this in the browser. And yes this will only get longer.
Full English FAQ will follow.

frontend/components/layout/NavBar.vue
```diff
@@ -110,6 +110,15 @@ function newRegistration(): void {
                             {{ $t("Studies") }}
                         </NuxtLink>
                     </li>
+                    <li>
+                        <NuxtLink
+                            to="/notes/FAQ"
```
EdoStorm96: Slugs are usually lowercase. I've also adjusted the fixture to reflect this.

```suggestion
                            to="/notes/faq"
```

EdoStorm96: Furthermore, Ideally, this slug wouldn't be hardcoded, but you'd have a query to the backend and wrap the navlink in a v-if that does not appear if no note with the slug faq exists. But this is maybe not necessary right now ...

frontend/locales/nl.json
```diff
@@ -117,7 +108,7 @@
     "Creating...": "Aan het aanmaken...",
     "Register new study": "Registreer een nieuwe studie",
     "Unknown study": "Onbekende studie",
-    "The study you are looking for could not be found.": "De studie die u zocht kon niet worden gevonden.",
+    "Oops ... The study you are looking for could not be found.": "Oeps ... De studie die u zocht kon niet worden gevonden.",
```
EdoStorm96: The Oopses are from an old branch (my work), but Xander removed this at some point I think.

frontend/pages/index.vue
```diff
@@ -21,41 +20,10 @@ definePageMeta({
                 <div class="uu-container">
                     {{
                         $t(
-                            "MyResearch is a website that helps researchers with submitting a request to register a research. On this website you can register your research at the processing registry. First, however, you should fill in the intake form",
+                            "MyResearch is a website that helps researchers manage their research administration/MyResearch is a website that helps researchers with submitting a request to register a research. On this website you can register your research at the processing registry which than can be seen by the privacy officer.",
```
EdoStorm96: So yeah I don't really like this text which is clearly unfinished. I am tempted to make a suggestion myself, but I guess we'll wait for Desiree to do so. I wouldn't want it to be merged like this.

EdoStorm96 (commented):
This is definitely a step in the right direction IMO! I like how it looks in general. I've regenerated the fixture to be more plain. I've also made some adjustments to the text, lie using dumb quotes instead of smart ones:

Image

This just keeps everything a bit neater in code.

I would wait till Desiree has a final version of the text, before merging this. I would also like to then have one more pass at reviewing.

frontend/locales/nl.json
RoordinkFrank: I saw it as well. I suspect it's a merge that had gone wrong. Will remove it.

RoordinkFrank:
> This is definitely a step in the right direction IMO! I like how it looks in general. I've regenerated the fixture to be more plain. I've also made some adjustments to the text, lie using dumb quotes instead of smart ones:
> Image
> 
> This just keeps everything a bit neater in code.
> 
> I would wait till Desiree has a final version of the text, before merging this. I would also like to then have one more pass at reviewing.

Thanks. This is the type of feedback I needed.

miggol:
I've rewritten the landing page text and translations.

Can you take another look at Edo's comments? For me, the FAQ link was not yet fixed. Also it would be nice if users who were not logged in could visit the FAQ, but if you want to make a separate issue for that that's cool too.

frontend/components/layout/NavBar.vue
RoordinkFrank: I don´t think that is necessary for now. There will basically always be a faq and it will have to be hardcoded in some way. Unless I am misunderstanding what you mean?

frontend/pages/index.vue
RoordinkFrank: rewritten by michael at this point.

RoordinkFrank:
> I've rewritten the landing page text and translations.
> 
> Can you take another look at Edo's comments? For me, the FAQ link was not yet fixed. Also it would be nice if users who were not logged in could visit the FAQ, but if you want to make a separate issue for that that's cool too.

the FAQ is a note and notes require no permissions. So that is exactly what is happening right now. You can visit the faq without loggin in.

EdoStorm96 (approved):
I've finished up some loose ends here. I think this is ready to merge!

PR #291, opened 2026-06-24
Repeatables!

miggol:
Herein lies the base framework for repeatable questions and steps.

### What to expect

When this PR is ready for review, this is what I want it to contain. Mostly, the backend part.

- [x] Repeating steps
- [x] Repeating questions
- [ ] Responses tied to specific repeats
- [ ] Conditions responding to specific repeats

### What not to expect

- [ ] Frontend management of repeats

The frontend aspects of this are somewhat beyond my abilities at this point, and it would be efficient to split up some of the development work, so I'm leaving that to whoever feels like picking up the strijdbijl.

## Outline

The core difference between a repeatable step or question and their regular counterparts is that they have a `ManyToMany` relationship with the new `RepeatIndex` model.

### Repeat indices

Where a repeat index would have previously been just an integer, it is now an object in the database. The repeat index model doesn't do much else. So really the integer has been upgraded to a pk. They are now unique and can be queried quickly through the ORM.

Besides their reverse link to repeatable questions and steps, repeat indices have two further attributes:

* Their parent (ForeignKey to RepeatIndex)
* Their linked submissions (ManyToMany to UFS)

#### Parent

A parent link is required for cases in which say, a repeatable question exists within a repeatable step. If multiple repeats of the step exist, and the user requests a new repeat of the question, the new RepeatIndex need to know to which repeat of the step it belongs. A parent attribute solves this problem. 

Repeatables need only link to one direct parent, that is enough to determine where everything goes. Repeats without a parent belong to top-level repeatable steps or questions.

#### Submissions

A repeatable object may have thousands of repeats spread across the database. But usually we're only interested in those connected to the current submission, which will at most be a couple. That's why there's a link to submissions.

Rather than a single ForeignKey, this is a ManyToMany field for the (future) case of revisions. For revisions, we want the following to be true:

* We can link repeats between revisions for the purposes of tracking changes and providing a diff. Even when the order or number of participant groups changes, we want to know which groups line up with others from historical revisions.
* When a revision removes a repeat, that must not affect previous revisions.

Therefore, repeat indices can be shared between revisions, and therefore be connected to multiple submissions. And like with attachments, before we delete them, we check if they're still connected to other submissions. If they aren't, we can delete them. But if they are still used elsewhere, we only unlink them from the current submission.

### Background instances

I've taken some liberty in this design decision even though I'm not taking charge of how this will work exactly on the Frontend side. So consider background instances as not at all set in stone. We can do this differently.

The issue that I'm trying to solve with background instances is that, if there is a repeatable step (or question) in the form, but there are no repeats yet, how does the frontend even know there's something that could be repeated there? And equally importantly, if the user wants to create a repeat for it, what ID do they pass along to the CreateRepeat mutation?

The answer for now is background instances. For every repeatable step or question in the form the backend always includes a background step or question. It is intended to not be displayed to the user, it takes no responses, and in the case of a step it has no substeps or questions. It merely exists to inform the frontend of the existence of a potential repeat together with its id.

### Mutations

For now, the only two mutations are `CreateRepeatMutation` and `DeleteRepeatMutation`. To create a repeat, you pass along the submission, the repeatable's id, and the parent RepeatIndex if applicable.

To delete a RepeatIndex, and therefore a repeat, all we need is its pk and the submission from which it should be deleted. As described above, if it is in use by other submissions, it will only be unlinked from the current submission rather than deleted.

Currently deletion does not delete all child repeat indices. We may consider to "hide" instead of delete repeats in the future, allowing for undo functionality.

### Responses

Responses and the interaction with conditions is still WIP. But it should be noted that even normal responses to non-repeatable questions can still carry a RepeatIndex. Because any kind of question can exist under a RepeatableStep, they need to be able to be saved with a RepeatIndex to assign them to the correct instance of the repeatable step.


XanderVertegaal:
(Hype!)

miggol:
Development continues in #371

PR #295, opened 2026-06-29
Update frontend dependencies

XanderVertegaal:
We still cannot update Nuxt, I'm afraid :( (so #258 cannot be closed), but at least we can hopefully close #293 after this.

Tested by doing a clean build:
``` 
docker image rm mr-vue-dev
docker compose build vue-dev
docker compose --profile dev up --watch
```

Update: I had to remove `package-lock.json` and `node_modules` as well.


frontend/plugins/fontawesome.ts
```diff
@@ -1,10 +1,14 @@
-import { library } from "@fortawesome/fontawesome-svg-core";
+import { library, config } from "@fortawesome/fontawesome-svg-core";
 import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome";
 
 import { faGear, faHouse } from "@fortawesome/free-solid-svg-icons";
 
+config.autoAddCss = false;
```
XanderVertegaal, line 6: This is based on https://docs.fontawesome.com/web/use-with/vue/use-with. It's not related to the dep upgrades, but I was struggling with the type error on line 13 when I found this guide. Apparently it's better to include the FontAwesome CSS globally in `nuxt.config.ts`.

frontend/nuxt.config.ts
```diff
@@ -23,6 +23,7 @@ export default defineNuxtConfig({
 
     vite: {
         server: {
+            allowedHosts: ["mr-vue"],
```
XanderVertegaal, line 26: For some reason, I need to add this now in order to get a working build.

frontend/package.json
```diff
@@ -17,51 +17,51 @@
...
+        "graphql": "^15.10.2",
+        "graphql-tag": "^2.12.7",
         "lodash": "^4.18.1",
         "lucide-vue-next": "^0.544.0",
         "nuxt": "3.21.1",
-        "pinia": "^3.0.3",
-        "uu-bootstrap": "git+https://github.com/UtrechtUniversity/UU-Bootstrap.git#v1.5.0-alpha.1",
-        "uuid": "^14.0.0",
+        "pinia": "^3.0.4",
+        "uu-bootstrap": "github:UtrechtUniversity/UU-Bootstrap#v1.5.0-alpha.1",
```
XanderVertegaal, line 39: This change was done automatically with `npm update --save`. It works for me; hopefully for you too!

EdoStorm96 (approved):
Looks fine to me :)

PR #297, opened 2026-07-01
Feature/delete study

### LINKED ISSUE #197: Delete Study
You should be able to delete studies in two places:

- On the study detail page
- In the study list (after every row)

Only studies whose status is DRAFT should be eligible for deletion. You should only be able to delete your own studies.

XanderVertegaal:
Closes #197 

Also includes commits from #295 (which should probably be "reviewed" first).


backend/research/models/study.py
```diff
@@ -100,7 +107,7 @@ def with_answers_by_annotation_key(self, queryset):
             answer = (
                 QuestionResponse.objects.filter(
                     submissions__study=OuterRef("id"),
-                    question_id=question.id,
+                    question_id=question.id,  # type: ignore
```
XanderVertegaal, line 129: This is simply to get rid of the red squiggly/satisfy the type checker.

backend/research/admin.py
```diff
@@ -32,14 +32,21 @@ def has_add_permission(self, request, obj=None):
...
+        "study_title",
+        "created_by",
+        "created_at",
+        "is_deleted",
+        "current_status",
+    )
     list_filter = ("created_at", "created_by")
     search_fields = ("id", "created_by__full_name", "created_by__email")
-    readonly_fields = ("study_name", "created_at")
+    readonly_fields = ("study_title", "created_at")
```
XanderVertegaal, line 45: This is an unrelated fix for the study page in Django Admin.

frontend/pages/studies/[studyId]/delete.vue
Comment (XanderVertegaal) [inline frontend/pages/studies/[studyId]/delete.vue:1]: With `useConfirm({...})` I don't think we need this file anymore.

frontend/components/studyDetail/AvailableActions.vue
EdoStorm96, line 1: Looking very nice ;p

backend/research/models/study.py
```diff
@@ -11,12 +11,19 @@
...
+            return queryset
+        return queryset.filter(created_by=user)
 
     def _editable_objects(self, user: User):
-        return self.filter(created_by=user)
+        return self.filter(is_deleted=False, created_by=user)
+
+    def _deletable_objects(self, user: User):
+        queryset = self.filter(is_deleted=False)
+        if user.is_privacy_officer or user.is_fetc_member:
```
EdoStorm96, line 26: I think I would prefer to not have the po/fetc delete items that have not been submitted ... As these will then get hard deleted. Might be a bit rough in some cases. You could also check in the mutation that only the creator is able to hard delete studies.

EdoStorm96 (approved):
Yep yep yep. Looks great and works well! I have one little request (to not make to po able to hard delete unsubmitted studies), but otherwise this is good to merge. I hope you get the chance to do so before your break :). Honestly, I'd rather see your changes to available actions get merged first, so I can go from there ;p

XanderVertegaal:
I'll get to it right away!

PR #298, opened 2026-07-01
Feature/study is seen

### LINKED ISSUE #211: Read/unread functionality for PO
Considering the minimal viable product may not include explicit review functionality. If that is the case, a flag that can mark a registration as seen would be much appreciated.

EdoStorm96:
fixes #211 

Made a little is_seen attribute for studies so the PO can mark stuff as seen/unseen. there is a StudyAction on detailpages and a little display item for lists and also a filter for lists.

I've also made changes to StudyActions in the frontend, they now do not necessarily need to redirect to a new page, but can also handle other stuff. In this case a popup. I know @XanderVertegaal is working on something similar, but we've coordinated it so that our changes are fairly compatible on this component, but some merge conflicts will ensue.

I've also modified create_dev_data so that 30% of studies it creates, will also receive a submitted StatusChange, to make testing easier. These will be filled in completely and 50% of the time receive is_seen = True.


frontend/pages/studies/index.vue
```diff
@@ -130,6 +138,19 @@ const filters = computed<UUListTypes.FilterDefinition[]>(() => {
         type: "checkbox",
         initial: [],
     };
+    const isSeenFilter: UUListTypes.FilterDefinition = {
+        field: "isSeen",
+        label: t("Seen by PO"),
+        options: [
+            ["true", t("Seen")],
+            ["false", t("unseen")],
```
miggol: Can you capitalize "unseen" ? :D

backend/research/utils/study_actions.py
```diff
@@ -75,3 +82,47 @@ class StudyDeleteAction(StudyAction):
...
+    """Base class for actions that require privacy officer permission."""
+
+    @classmethod
+    def _can_perform_action(
+        cls,
+        study,
+        user,
+    ) -> bool:
+        if study.status == SubmissionStatus.DRAFT:
+            return False
```
miggol, line 99: Is this intended to be here? Seems unrelated to the PO base class.

backend/research/utils/study_actions.py
```diff
@@ -75,3 +82,47 @@ class StudyDeleteAction(StudyAction):
...
+
+    action = ActionEnum.MARK_SEEN_ACTION
+
+    @classmethod
+    def is_available(cls, study, user):
+
+        if study.is_seen:
+            return False
+
+        return cls._can_perform_action(study, user)
```
miggol: I think a more common pattern would be to call `super().is_available()` here. That allows for chaining of mixins, i.e. inheriting from both `PrivacyOfficerStudyAction` and a hypothetical `SubmittedStudyAction` that checks if a study is submitted.

frontend/components/studyDetail/AvailableActions.vue
```diff
@@ -55,6 +59,47 @@ const slug = computed<string | null>(() => {
...
+                messages
+            }
+        }
+    }
+`);
+
+const { mutate: markStudySeen } = useMutation(MARK_STUDY_SEEN_MUTATION);
+
+function checkMarkStudySeen(isSeen: boolean): void {
+    useConfirm({
```
miggol: Considering this is an easily reversible action, I don't think `useConfirm` is necessary here.

frontend/pages/studies/[studyId]/index.vue
```diff
@@ -56,6 +60,14 @@ const studyStatus = computed(() =>
                 <div class="uu-container">
                     <div class="row">
                         <div class="col me-5">
+                            <span
+                                v-if="
+                                    currentUserStore.currentUser
+                                        ?.isPrivacyOfficer && study.isSeen
```
Comment (miggol) [inline frontend/pages/studies/[studyId]/index.vue]: I think it's better to have the unseen badge visible on this page as well.

Additionally, I think this information should be available or not depending on if the backend sent it. We don't trust the frontend, so it shouldn't do the checking of PO status.

backend/research/types/StudyType.py
```diff
@@ -82,6 +99,7 @@ class Meta:
             "created_at",
             "reference",
             "title",
+            "is_seen",
```
miggol: A potential easy fix for not sending seen info to non-PO users would be to just `resolve_is_seen` to true, false, or unknown. Normal users only get unknown.

miggol (changes requested):
These features are really coming together fast!

A bit of minor subjective stuff. But I do have one requested change which is to not provide `is_seen` info when the user is not a PO. Or provide "unknown".

backend/research/utils/study_actions.py
EdoStorm96: I guess for now, this is required for both actions which inherit from the baseclass (also return to draft which is coming) so I'd be open to leave it in.

backend/research/utils/study_actions.py
EdoStorm96: Yes that makes perfect sense indeed lol

frontend/pages/studies/[studyId]/index.vue
```diff
@@ -56,6 +60,18 @@ const studyStatus = computed(() =>
                 <div class="uu-container">
                     <div class="row">
                         <div class="col me-5">
+                            <div v-if="study.isSeen !== null">
+                                <span
+                                    v-if="study.isSeen"
+                                    class="badge rounded-pill text-bg-info mb-1 fs-6"
+                                    >Seen</span
```
Comment (miggol) [inline frontend/pages/studies/[studyId]/index.vue]: These two are untranslated

miggol (approved):
Nice! Just found some untranslated strings, otherwise ready to go

PR #301, opened 2026-07-07
Feature/return to draft + Progressbar

EdoStorm96:
So now we can return submitted studies back to draft!

For testing I would suggest you make your form very short, so you can easily fill it in. This was my setup (you will need faculty, because that has an annotation key.):
image

Then, if you are a PO (eg. admin) you can return submitted studies back to draft.

As a bonus, I've updated the progressbar to reflect actual statuschanges. There are still some fake bubbles (created and a disabled submitted if the status is draft) so we don't end up with a progressbar with only a draft item. I like how it looks for now.


backend/research/mutations/CreateStatusChangeMutation.py
```diff
@@ -0,0 +1,46 @@
+from graphene import Field, ID, List, Mutation, ResolveInfo, Argument
+from graphene_django.types import ErrorType
+
+from main.models import MRPermission, User
+from research.models.study import Study
+from research.models.reviews import StatusChange, SubmissionStatus
+from research.types.StudyType import GQLSubmissionStatus, StudyType
+
+
+class CreateStatusChange(Mutation):
```
miggol: I agree with copilot about the permissions here, but I would suggest that we make this a mutation specifically for returning to draft.

I wouldn't want to have one endpoint for all the logic and checks we that might ever want to do relating to status changes. Making these mutations specific allows you to drop the status argument as well.

frontend/components/studyDetail/StudyProgressBar.vue
```diff
@@ -1,110 +1,96 @@
...
+    return [
+        // Add a "fake" created StatusChange at the beginning
+        {
+            label: t("Created"),
+            createdAt: null,
+            isComplete: true,
+            isDisabled: false,
+        },
+        ...items,
+    ];
```
miggol: It would probably look better if the created datetime would be displayed here in the created step. This just feels off:

Image

It's hard to find a perfect solution, and this is definitely fine for now. But a slightly better solution IMO would be as follows:

* Created _jan 18_
* Draft _now_
* (grey) Submitted

So the rule would be:

Always hardcode the first status change to display "Created" and display its date.

Display all returns to draft by PO as "Returned to submitter" (which I think is more informative from a researcher's POV) with the date

If a study is a draft, always append the following two steps

* Draft _now_
* (grey) Submitted

miggol (changes requested):
Works well for me! The only change I would like to see before merge would be the mutation. The rest can be future improvements.

There are also a bunch of missing frontend translations but we can just address them at the end of this week in one big swoop.

frontend/components/studyDetail/StudyProgressBar.vue
```diff
@@ -1,110 +1,96 @@
...
                             incomplete:
                                 !progressItem.isComplete &&
-                                progressItem.isActive,
+                                !progressItem.isDisabled,
                         }"
                     ></span>
-                    <span class="lh-1">{{ $t(progressItem.label) }}</span>
+                    <span class="d-flex flex-column">
+                        <span class="lh-1">{{ $t(progressItem.label) }}</span>
+                        <span
```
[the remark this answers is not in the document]
EdoStorm96: Yes, but is doesn't  work without it, so I 'll just leave it.

frontend/components/studyDetail/StudyProgressBar.vue
```diff
@@ -1,110 +1,105 @@
...
+    }));
+    // Add a "fake" Draft bubble when we only have one status change ("Created")
+    if (statuses.value.length === 1) {
+        items.push({
+            label: t("Draft"),
+            createdAt: t("Now"),
+            isComplete: false,
+            isDisabled: false,
+        });
+    }
```
miggol, line 59: My initial suggestion was to just always add the draft bubble if a study was a draft, like the submitted bubble. But this works too!

backend/research/mutations/CreateStatusChangeMutation.py
```diff
@@ -0,0 +1,57 @@
...
+                .get(pk=study_id)
+            )
+        except Study.DoesNotExist:
+            return cls(
+                ok=False,
+                errors=[ErrorType(field="study_id", messages=["Study not found."])],
+            )
+
+        # Only allow privacy officers to return SUBMITTED studies back to DRAFT
+        if not user.is_privacy_officer or study.status != SubmissionStatus.SUBMITTED:
```
miggol: I find this double negation with an OR harder to parse than 

```python
if not (user.is_privacy_officer and study.status == SubmissionStatus.SUBMITTED):
```

But maybe that's just me, feel free to ignore

backend/research/mutations/CreateStatusChangeMutation.py
```diff
@@ -0,0 +1,57 @@
+from graphene import Boolean, Field, ID, List, Mutation, ResolveInfo
+from graphene_django.types import ErrorType
+
+from main.models import MRPermission, User
+from research.models.study import Study
+from research.models.reviews import StatusChange, SubmissionStatus
+from research.types.StudyType import StudyType
+
+
+class CreateDraftStatusChange(Mutation):
```
miggol, line 10: I'm of two minds about this mutation name. It's very accurate but also slightly confusing... Do with this what you will

frontend/components/studyDetail/StudyProgressBar.vue
```diff
@@ -1,110 +1,105 @@
...
+        // Add the createdAt as the date, except if last status is draft, then just say "Now"
+        createdAt: useLocalDateTime(change.createdAt) || null,
+        // Submitted is always completed, the last item is always incomplete
+        // and the first ("Created") will always be complete
+        isComplete:
+            change.status === SubmissionStatus.Submitted ||
+            index < statuses.value.length - 1 ||
+            index === 0,
+        isDisabled: false,
+    }));
```
miggol, line 50: Good comments, nice code!

miggol (approved):
Approved! I had some comments but they're all optional

PR #304, opened 2026-07-09
Feature/overview page

EdoStorm96:
Well here's our final(ish) MVP feature! fixes #300 

We now have a view overview study action. This PR has also altered the normal overview step within the form, so be sure the have a look at that. I've just re-used the overview step. To be able to reuse it without making any alterations to things like useFormState, I've opted to use a design which basically mirrors how our Form works in the frontend. So we have the view.vue page, which has a component called the OverviewWrapper, which has the SubmissionOverview component. This is a bit silly, but otherwise we would have had to make changes to other things and this controls the flow of data nicely so the component never get surprised by the data types they are receiving.

I've made the overview action available for anyone who has view permission, so it it mostly available. Thanks @XanderVertegaal for helping me out with some stuff :) As I described, I basically was able to undo the changes to useFormState, by just more closely mirroring how the Form works. I think I prefer this to opening up a useFormState can of worms.

I removed the NuxtLink on study action tiles, as I thought that might have caused #303, but it appears the bug persists ... Please check if you encounter this. Anyway, the nuxtlink was not really doing anything anymore, so good riddance.

Let me know what you think and let's try and merge this today!


frontend/pages/studies/[studyId]/[submissionId]/view.vue
```diff
@@ -0,0 +1,187 @@
...
+    <div class="uu-content">
+        <Title>{{ $t("Registration overview") }}</Title>
+        <div class="uu-hero">
+            <h1>{{ $t("Registration overview") }}</h1>
+        </div>
+        <div class="uu-container">
+            <div class="col-12">
+                <h1>
+                    {{ study?.reference }} - <em>{{ study?.title }}</em>
+                </h1>
```
Comment (miggol) [inline frontend/pages/studies/[studyId]/[submissionId]/view.vue]: Another "go back" button at the top would be nice.

If you could wrap the study title and back button in a "sticky-top" bootstrap div, then they will always remain visible once you scroll past them. But that would be the 120% effort solution at this point. Just another button at the top is also fine.

frontend/components/studyDetail/AvailableActions.vue
```diff
@@ -145,7 +156,7 @@ const availableActions = computed<AvailableAction[]>(() => {
     <div v-else-if="availableActions.length > 0">
         <h3 class="mb-3">{{ $t("Available actions") }}:</h3>
         <div class="tiles">
-            <NuxtLink
+            <div
                 v-for="(action, index) in availableActions"
                 :key="index"
                 :style="action.style"
```
miggol, line 275: If you could add "cursor: pointer;" to the universal action style that would make them more obviously clickable. Now the cursor transforms to text select whenever you go over text, bit messy. Very minor obviously.

frontend/components/studyDetail/AvailableActions.vue
```diff
@@ -101,6 +99,19 @@ function deleteStudyWithConfirmation(): void {
 }
 
 const actionMap = computed<Record<ActionEnum, AvailableAction>>(() => ({
+    [ActionEnum.ViewAction]: {
+        label: t("View overview"),
```
miggol: IMO the label for this action should just be View/Bekijk, without "overview". And I must say these tiles look really nice with the icons.

backend/research/utils/study_actions.py
```diff
@@ -75,3 +76,11 @@ class StudyDeleteAction(StudyAction):
...
     def is_available(cls, study, user):
         return study in Study.objects.accessible_objects(user, MRPermission.DELETE)
+
+
+class StudyViewAction(StudyAction):
+    action = ActionEnum.VIEW_ACTION
+
+    @classmethod
+    def is_available(cls, study, user):
+        return study in Study.objects.accessible_objects(user, MRPermission.VIEW)
```
miggol, line 145: Love that adding one of these is so easy.

I think it would be okay to have the "View" option unavailable when a study is still a draft. Just to avoid confusion about where to go as a submitter. Not a strong opinion.

frontend/pages/studies/[studyId]/[submissionId]/view.vue
```diff
@@ -0,0 +1,187 @@
...
+                        }
+                        ... on DateQuestionType {
+                            futureOnly
+                        }
+                    }
+                }
+            }
+        }
+    }
+`);
```
Comment (miggol) [inline frontend/pages/studies/[studyId]/[submissionId]/view.vue]: Are each of our GET_FORM queries notably different? Would be so much nicer to import them from somewhere, and just have one. Hard repeating it is bound to go wrong when we add a Question type or something.

miggol (approved):
Very happy to see #303 fixed as well.

I can't say that I completely understand the UseFormState situation, but on a high level I find the use of a Wrapper here very appropriate and I like how little needs to be changed to re-use the overview step. Great work.

Approved as I wouldn't consider anything here blocking. But adding the extra button at the top and changing the action tile label would be my top recommendations.

frontend/components/form/overview/OverviewWrapper.vue
```diff
@@ -0,0 +1,36 @@
...
+    },
+);
+
+void v$.value.$validate();
+
+watchEffect(() => {
+    if (formObject.value) {
+        useAnnotateErrors(v$.value, formObject.value);
+    }
+});
```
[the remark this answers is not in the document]
EdoStorm96: It works fine, so I will ignore this.

backend/research/utils/study_actions.py
EdoStorm96: I do see your point ... But I'm gonna leave it for now. Now the PO can also always see it, which I think is desired?

frontend/pages/studies/[studyId]/[submissionId]/view.vue
Comment (XanderVertegaal) [reply frontend/pages/studies/[studyId]/[submissionId]/view.vue]: This is definitely possible, using [fragments](https://www.apollographql.com/docs/react/data/fragments). I tried setting this up for MR some time ago and got close to a working solution, but I ran into type generation issues and had to move on to more pressing matters. I'd be happy to give this another look though!

PR #309, opened 2026-07-10
Restore frontend volume and modify i18n-extract command

miggol:
Working on translations I ran into difficulties moving files back from the container to disk. This should alleviate that while keeping the benefit of watch functionality.

Also, by default `vue-i18n-extract` uses periods as a seperator in translation keys. That doesn't work for those who choose to live dangerously and use full strings as translation keys, because it splits up any text consisting of more than one sentence. So I chose a random separator that hopefully (you never know!) will never come up naturally.


XanderVertegaal (commented):
I'm not sure I understand. What problem were you running into exactly? It seems like this PR basically removes the watch functionality (apart from changes in `package.json` and `vite.config.js`, which are unlikely to change anyway).

The old issue returns: if I start the container with a type error, I cannot get the error message to go away (even after fixing the error itself). To reproduce, replace `{{ useTranslateableAttribute(step, "name") }}` with `{{ useTranslateableAttribute(step, 4) }}` in OverviewForm.vue and start the container. Then put `"name"` back again; the type error will not go away. This is what the watch functionality solves.

Let's look at this together sometime this week!

I still think we should consider reverting back to using translation keys instead of living dangerously, but obviously now is not the time to discuss/do this. Using a random separator string is fine.

EdoStorm96 (commented):
I am not fully grasping the consequences of this PR, but I did give it a quick whirl.

@XanderVertegaal, I have attempted to reproduce what you have described, but for me the error does go away if I fix it in my source code. I did run my containers with --watch. Does that maybe make a difference?

PR #302, opened 2026-07-07
fix: backend validation edo's feedback

RoordinkFrank:
I did merge to soon.

PR #313, opened 2026-07-15
Fix/backend validation handling + other minor issues

EdoStorm96:
fixes #311 

Backend validation was not working properly. This is fixed now. The backend now just sends an error to the response of the graphql request. When a response does not pass validation, it does not get saved. However all other responses do get saved.

I've also made a pretty major change to how we submit responses (which was actually pretty minor to implement). Instead of submitting the whole form everytime, we now only submit the responses for the current step. This should save a lot of time on the backend as we do not validate the whole form on every update_submission, just the step that has been edited. This was actually only a change to the frontend. The backend was totally ready for this and did not require any changes to implement this, so that is nice.

I've also applied the fixes from #302 here, so I closed that branch, as it was no longer relevant.

### DIFF
--- a/backend/form/mutations/UpdateUserFormSubmission.py
+++ b/backend/form/mutations/UpdateUserFormSubmission.py
@@ -31,7 +31,7 @@ class UpdateUserFormSubmission(Mutation):
         user: User = info.context.user
 
         try:
-            submission = update_submission(
+            submission, errors = update_submission(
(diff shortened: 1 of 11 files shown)


EdoStorm96:
Also implemented fix to #307

EdoStorm96:
Also implemented a fix for #306

PR #318, opened 2026-07-17
Fix  bring fixture in line with server

EdoStorm96:
During my demo to our tester, they requested I add one more question to the form. I did this live on the server, but here it is reflected in our fixture.

I also found that MRDocument was not yet excluded from our fixture generator so I did that as well.

I did not have the chance to update the excel with this change before my vacation. @RoordinkFrank Could you take this up?

### DIFF
--- a/backend/form/fixtures/vwr.json
+++ b/backend/form/fixtures/vwr.json
@@ -222,6 +222,23 @@
     },
     {
         "model": "form.basequestion",
+        "pk": 61,
+        "fields": {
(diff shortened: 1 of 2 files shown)


RoordinkFrank (approved):
Looking good. The rest of the VWR had "E.g.," instead of "E.g." That's one thing Desiree will comment about once she is back. That said probably easier to add that in the next version then fixing it in the live server again and generating new fixtures.

PR #319, opened 2026-07-20
feat/bring-back-stepinfoquestion

RoordinkFrank:
fixes #317 

Wants to merge into fix--bring-fixtures-in-line-with-server. I did not want to have a fixture merge conflict.

--------------
commit accordion spaced v1
image
Additional information is small here, whitespace left on the right. If we want to keep michaels design where the step description is above the additional information this seems to the best way to me. Makes the yellow also a bit less overwhelming.


frontend/components/form/FormWrapper.vue
```diff
@@ -251,7 +251,7 @@ function navigateToSlug(slug: string) {
             :stepper-config="formStepperConfig"
             @step-clicked="navigateToSlug"
         />
-        <div class="col-12 col-lg-9">
```
RoordinkFrank, line 254: This made the forms + now info larger towards the right side of the screen, problem is that it was overriding the parent uu components which sets all the text in the middle. I assume it was actually used in the past to make the text less broad, regardless it's not needed at this point.

backend/form/forms.py
```diff
@@ -114,6 +122,12 @@ def __init__(self, *args, **kwargs):
                     map(str, current_substep_order)
                 )
 
+            current_step_info_order = self.instance.get_stepinfo_order()
```
RoordinkFrank, line 125: I did not know this getter was autogenerated.

backend/form/models/form.py
```diff
@@ -66,8 +66,10 @@ def __str__(self):
         return f"{self.name} ({self.pk})"
 
 
-class StepInfoText(models.Model):
-    step = models.OneToOneField(
-        Step, on_delete=models.CASCADE, related_name="info_text"
-    )
-    text = models.TextField()
+class StepInfo(models.Model):
```
EdoStorm96, line 69: Ok, so this is super annoying, but I actually prefer the name StepInfoText for this case. I think Xander suggested StepInfo as a name when we just had one per step. We had the step and its info. Now that there are multiple "info's" per step, this name becomes very ugly because "info" or "information" is an [uncountable noun](https://ieltsliz.com/uncountable-nouns-word-list/). info's or informations is always incorrect.

So now that we do have multiple _StepInfoText(s)_, I would say `StepInfoText` is the superior name ...

frontend/components/form/FormSideBar.vue
```diff
@@ -1,39 +1,75 @@
...
         </div>
     </div>
 </template>
 
-<style lang="scss" scoped>
-.help-item:not(:last-child) {
-    margin-bottom: 2rem;
-}
-</style>
+<style lang="scss" scoped></style>
```
EdoStorm96, line 75: This can go right?

EdoStorm96 (changes requested):
Really nice work! This was very nice to find returning from my break!

I think in general this looks great. I do have two things ...

So first, I don't like the name StepInfo here. I think it should be StepInfoText or StepInfoQuestion. I explain this more in my other comment. At the end of the day, if it is not worth you time, I do get that ... And I can live with that. But if you can fix it, I think another name would be better, because info is an uncountable noun. See my other comment for more reasoning about this haha.

Secondly, I think the design can be improved. I've gone ahead and implemented my vision in #336. If you merge my PR into here, I think this is good to go! (Though I would really appreciate changing the name of stepinfo ...)

PR #333, opened 2026-07-27
Align middle study list table data

evania:
Closes #310 
All table data columns in the study list are now aligned to the middle (vertical alignment)


EdoStorm96 (approved):
Yep, this looks all good! Nice job 💫

PR #334, opened 2026-07-28
fix: restrict the use of _delete_document_if_cleared

EdoStorm96:
fixes #331

this was caused by _delete_document_if_cleared. It now only runs when the new answer is actually an empty string and the question is a FileUploadQuestion.

### DIFF
--- a/backend/form/services/update_submission.py
+++ b/backend/form/services/update_submission.py
@@ -80,7 +80,7 @@ def update_submission(
                     )
                     new_response.submissions.add(current_submission)
                 else:
-                    if not response["answer"].get("value") and isinstance(
+                    if response["answer"]["value"] == "" and isinstance(
                         qr.question.get_subclass(), FileUploadQuestion
                     ):
                         _delete_document_if_cleared(qr.answer, response["answer"])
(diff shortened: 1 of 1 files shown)


backend/form/services/update_submission.py
```diff
@@ -79,8 +80,11 @@ def update_submission(
                     )
                     new_response.submissions.add(current_submission)
                 else:
+                    if not response["answer"].get("value") and isinstance(
```
miggol: To be exact, you're checking for a falsey value here, and then again in `_delete_document_if_cleared()`.

I think the falsey check on this line is unnecessary. It will also trigger on int(0), False, None, empty list, etc. so it's not that helpful. Just checking the question type is fine here.

miggol (approved):
This does not address issue #331 as I understand it, please unlink so it doesn't close the issue. But I can see what you're trying to fix. Approved with a comment

backend/form/services/update_submission.py
RoordinkFrank: #331 will be fixed in this branch. Edo is correct here.

RoordinkFrank (approved):

PR #344, opened 2026-07-30
Feat/small issues post testing

### LINKED ISSUE #328: add study title to form
Image

### LINKED ISSUE #340: Study action labels change
Make it only whole verbs:
Bekijken - Terugzetten naar concept - Markeren als gelezen - Verwijderen
View - Return to draft - Mark as seen - Delete

### LINKED ISSUE #341: Inform users their progress is continually saved
Our PO wants users to be reminded that their progress is constantly saved. I am suggesting to do this using a notification whenever you go to a new step in the form. It is a little annoying to me ... But also reassuring i guess? And very easy to implement.

EdoStorm96:
fixes #328 
fixes #341 
fixes #340 

This PR fixes a few small-ish issues that came up after testing.

The title of a study is now always displayed in the hero of a form (and as the tab title). I've also repurposed this design in the hero of our view page, as that was still pretty ugly.

Secondly, our tester wanted to make it very clear to users when their progress is saved. I opted for a specific save button, that is always available (although it is actually save more often, but I think this is fine.) Related to this, our tester also wanted the submit button to be disabled if the form still contained errors. This is now also the case.

The labels for our StudyActions needed to be changed. This was just a fix to nl.json.


RoordinkFrank (changes requested):
Submit button works great. You worked quick with that.
Renamed action buttons are as expected.
Title changes are here as well. They will not change live until saved but that is more than acceptable.

The save and go back button works great. I do like the notification. What I do not like however is the location. I want the MR-26-0002 - title to be in the place where the button is now and I want the button to be with the other buttons. That seems a way more logical place for me. The black color of the button can stay, makes it stand out more that you are about to navigate away.

frontend/pages/studies/[studyId]/[submissionId]/view.vue
```diff
@@ -1,67 +1,45 @@
...
+        >
+            <BSButton
+                variant="secondary"
+                class="btn-arrow-left flex-shrink-0"
+                @click="goBackToStudy"
+            >
+                {{ $t("Go back to study page") }}
+            </BSButton>
+            <h1 class="mb-0 text-wrap text-break">
+                {{ study?.reference }} -
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[studyId]/[submissionId]/view.vue]: Good idea to use the hero bar for the study title! That feels very natural.

I'm not a fan of having the title pushed to the right by the 'back' button. Can you put the 'back' button below the hero or at least put the button and the study title on separate lines? The 'back' button itself may be a bit less prominent as well. (Just a single line of underlined text would work, I think.)

frontend/composables/useUpdateFormSubmission.ts
```diff
@@ -0,0 +1,40 @@
...
+            ok
+            errors {
+                field
+                messages
+            }
+        }
+    }
+`);
+
+export function useUpdateFormSubmission(shouldEvictStudy: boolean = false) {
```
XanderVertegaal: The type annotation is not necessary; TS can infer it based on the default value. Also, it may be a little bit neater to specify what the parameter does, so something like `reloadStudy` is more informative. (The fact that you reload a study by doing a cache evict is an implementation detail.)

```suggestion
export function useUpdateFormSubmission(reloadStudy = false) {
```

frontend/composables/useStudyQuery.ts
```diff
@@ -0,0 +1,24 @@
...
+    query GetStudyTitle($id: ID!) {
+        study(id: $id, mrPermission: "View") {
+            id
+            title
+            reference
+        }
+    }
+`);
+
+export function useStudyQuery(studyId: Ref<string | undefined>) {
```
XanderVertegaal: Maybe `useStudyTitleQuery`? I feel we might risk a clash with other study queries in the future 😆

frontend/components/form/FormWrapper.vue
```diff
@@ -53,6 +60,9 @@ const { mutate: mutateForm } = useMutation<UpdateUserFormSubmission>(
     {
```
XanderVertegaal, line 62: Don't you want to replace this with your new `useUpdateFormSubmission` composable?

frontend/components/form/FormWrapper.vue
```diff
@@ -108,6 +117,10 @@ function submitForm(options = { submit: false }): void {
         });
 }
 
+defineExpose({
```
XanderVertegaal: This is a clever solution, but I generally avoid using `defineExpose` unless it's necessary, since it makes it hard to track where this function is called from, and blurs the distinction between the child and parent component a bit. It feels a bit like an anti-pattern.

If you need it, I would rather outfactor this into a shared composable instead.

frontend/components/form/FormWrapper.vue
```diff
@@ -295,7 +310,15 @@ function navigateToSlug(slug: string) {
                     {{ $t("Previous") }}
                 </BSButton>
                 <BSButton
-                    v-if="selectedStep.isOverview"
+                    v-if="showSubmissionWarning"
+                    :disabled="true"
```
XanderVertegaal: Fun fact: instead of `:disabled="true"` you can also simply add `disabled` (without the `:`).

XanderVertegaal: There are two submit buttons in the template right now. Is it possible to merge them to make things a bit DRYer?

frontend/pages/studies/[studyId]/[submissionId]/[slug].vue
```diff
@@ -1,25 +1,70 @@
...
+const study = useStudyQuery(studyId);
+const { t } = useI18n();
+
+const formWrapperRef = ref<{ submitForm: () => void } | null>(null);
+
+function handleBackNavigation() {
+    // We call submitForm from our FormWrapper component
+    formWrapperRef.value?.submitForm();
+
+    useNotification(t("Your progress has been saved."), "info", 3);
```
Comment (XanderVertegaal) [inline frontend/pages/studies/[studyId]/[submissionId]/[slug].vue]: I'm not seeing these notifications pop up, but I'm actually glad: it sounds like a lot of notifications. I'd rather have some static text at the top that indicates that all changes are saved. In Lettercraft (an RSL project) we have an indicator that marks that everything is saved.

 (This changes to "Saving in progress" when a mutation is happening, but that is probably overkill for now.)

Something similar at the top of the form could work here too (and is simpler than the `defineExpose` approach, I think).

XanderVertegaal (changes requested):
Quick work! I really like the use of the hero bar for the study title. <3

The notifications don't seem to trigger for me, but I'm also not sure I'd be in favour of that approach. Would a static notification/bar suffice as well? Then we don't need the `defineExpose`.

frontend/components/form/FormWrapper.vue
```diff
@@ -242,9 +256,27 @@ function navigateToSlug(slug: string) {
...
+
+function handleBackNavigation() {
+    submitForm();
+    useNotification(t("Your progress has been saved."), "info", 3);
+
+    return navigateTo({
+        name: "studies-studyId",
+        params: { studyId: studyId.value },
+    });
+}
```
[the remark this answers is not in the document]
EdoStorm96: I am not too worried about this, because in reality we are constantly saving. This is just an extra assurance for users, but not really the main saving functionality.

RoordinkFrank (approved):

PR #346, opened 2026-07-31
hotfix: bring back nav buttons on overview step

EdoStorm96:
So I messed up in #336 and pushed a bug to develop. The overview step and normal steps of the form are now in their own v-if v-else block and I forgot to also implement the buttons for the overview step block. Sorry about that ... :sweat_smile:


XanderVertegaal (approved):
Works well!

PR #364, opened 2026-08-18
Feat/repeatable step component

EdoStorm96:
So here is a frontend component that allows for the creation/management of repeatable steps! Yay!

How I've gone about this is to create a `RepeatableStepQuestion`. This is just a new question type, that does have some quirks ... Unlike other questions, there is no answer provided by the user here. It is only used to trigger mutations. I did want this question to be aware of the repeats that are currently being managed by it, so I've repurposed the question's answer for this. Its answer is a list of repeat indices, that is being managed purely by the mutations. This answer is then used in the frontend to query the repeats. It worked out quite conveniently imo.

I did run into some (unrelated) bugs here and there, which I've fixed along the way in order to make my feature work properly. But now this really seems to work great. I can even save data in repeated steps! Thanks to the great work in the backend by @miggol. This was kindoff a breeze :)

I´ve discussed with @miggol that he background steps have become redundant with the `RepeatableStepQuestion`. However, he told me how would properly remove this in the main branch.

Let's get this merged soon.

Little sneak peek at the component:
image

Note: The text here shows the repeat indices. I think as we're still testing this feature, this is useful, but this will of course get removed before merging to develop.

To test this, create a clean DB. run `python manage.py load_fixtures --users-only` and start to create a form, with a step with a `RepeatableStepQuestion` that links to a repeatable step and start having fun.


frontend/components/form/RepeatableStepQuestion.vue
```diff
@@ -0,0 +1,218 @@
...
+
+interface Props {
+    question: RepeatableStepQuestionWithValue;
+    isInvalid: boolean;
+}
+
+const props = defineProps<Props>();
+const emit = defineEmits<{
+    (e: "repeat-step-clicked", slug: string): void;
+}>();
```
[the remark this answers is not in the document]
EdoStorm96: I don't think this is really causing any issues? So I'm going to leave this for now.

backend/form/mutations/repeat_mutations.py
```diff
@@ -55,6 +63,26 @@ def mutate(
...
+            return cls(errors=[error])
+
+        # Use a QuestionResponse to store
+        if response_id is not None:
+            qr = QuestionResponse.objects.get(
+                pk=response_id,
+            )
+        else:
+            qr = QuestionResponse.objects.create(question=rsq, answer={"value": []})
+            qr.submissions.add(submission)
```
miggol: Hmm, I can see how this fits into the pieces that we already have. But I'm hesitant to actually creating a question response per repeat. Although unlikely to happen under natural circumstances, it does create the possibility of desynchronization between the responses listed here and the repeat indexes that actually exist. 

Say if we ever had to edit a form that was in use, or if we would want to add a "delete this repeat" button elsewhere in the form (for file uploads perhaps?). 

However I do think it's a great idea to use the responses field for this. Great find. But if we could somehow insert these responses at resolve time (dynamically) rather than having them actually exist in the database I would prefer that.

backend/form/mutations/repeat_mutations.py
```diff
@@ -99,13 +132,27 @@ def mutate(
...
 
+        try:
+            qr = QuestionResponse.objects.get(
+                pk=response_id,
+            )
+            # Remove the repeat_id from the answer value list
+            answer_value = qr.answer.get("value", [])
+            repeat_id = int(repeat_id)
+            if repeat_id in answer_value:
+                answer_value.remove(repeat_id)
```
miggol: In the future keep in mind that if you do two interconnected things in the ORM that must happen atomically, i.e. they must both succeed or both fail, it's wise to wrap them in a transaction.

This is of lesser importance considering my comment above of course.

backend/form/models/questions.py
```diff
@@ -97,11 +97,13 @@ def get_subclass(self):
             "numberquestion",
             "datequestion",
             "fileuploadquestion",
+            "repeatablestepquestion",
         ]:
             if hasattr(self, subclass_name):
                 cls = getattr(self, subclass_name)
                 if hasattr(cls, "repeatable" + subclass_name):
                     return getattr(cls, "repeatable" + subclass_name)
+                return cls
```
miggol: Thumbs up

backend/form/services/user_form_resolver.py
```diff
@@ -81,14 +82,19 @@ def _make_step_repeats(
...
+        # remove these in the main branch. (18/08/26)
+
+        # # For repeatable steps, we first insert the "blank" background
+        # # step into the output.
+        # background_step = self._create_step_instance(
+        #     step,
+        #     repeat_index=parent_index,
+        #     background=True,
+        # )
+        # repeats.append(background_step)
```
miggol: Alright, I'm back up to speed now so you have my blessing to remove this yourself if it's in the way. Same for the background questions. Otherwise I'll get to it (and fix the tests) once I'm back from my holiday.

backend/form/mutations/repeat_mutations.py
miggol: The more I think about this the more I see advantages to fetching the existing repeats at resolve time. The form or question resolver could pass along info like the data subject group name. It would be quite unwieldy update this question's response every time a name changes.

backend/form/models/questions.py
```diff
@@ -167,6 +169,18 @@ class FileUploadQuestion(BaseQuestion):
...
 
+class RepeatableStepQuestion(BaseQuestion):
+    """
+    A special question, with which repeatable steps can be created, deleted and managed.
+
+    Does not have a response, but is used to trigger mutations.
+    """
+
+    repeatable_step = models.OneToOneField(
+        "form.RepeatableStep", on_delete=models.CASCADE
```
miggol: Is there anything stopping us from just linking to the base `form.Repeatable` class? Might be able to kill two birds with one stone and manage question repeats as well.

Actually I just thought of something, without custom code the django-admin experience would be not-so-friendly. So don't worry about it for now. But once we get round to the `RepeatableQuestionQuestion` (lol) we will have to do either that **or** otherwise revise the inheritance structure so we don't have to make separate RQQs for all the repeatable select, text, number, etc questions.

miggol: Also would be cool to have custom button texts as fields on this model at some point, with a generic default. Just thinking out loud.

backend/form/queries.py
```diff
@@ -80,3 +82,64 @@ def resolve_select_question(
...
+            queryset = UserFormSubmission.objects.accessible_objects(
+                user, MRPermission.EDIT
+            )
+
+            if not question_response.submissions.last() in queryset:
+                return []
+
+            # Extract repeat indices from the answer
+            answer = question_response.answer
+            repeat_indices = answer.get("value", []) if isinstance(answer, dict) else []
```
miggol: In the end, a repeat index is nothing but a `pk` linked to:

1. A repeatable
2. A list of submissions
3. (optional) A parent index

So if we have a repeatable and a submission, we should be able to get a list of indices without reading them from a response. It's the most up-to-date list of indices available.

Taken from `UserFormResolver._fetch_repeat_indices()`:

```python3
# Given that we have a repeatable object
indices = repeatable.repeat_indices.filter(
    submissions=submission,
)
# Then filter for correct parentage. This is only relevant
# for repeats within repeats.
if parent_index is not None:
    indices = indices.filter(parent=parent_index)
return indices
```

Maybe I'm missing something, but the whole response part still confuses me. Could we do without it?

Please keep the permission check though :)

frontend/components/form/RepeatableStepQuestion.vue
```diff
@@ -0,0 +1,218 @@
...
+
+                    <div class="d-flex gap-2">
+                        <button
+                            class="btn btn-primary"
+                            @click.prevent="
+                                emit(
+                                    'repeat-step-clicked',
+                                    `${step.slug}.${step.repeatIndex}`,
+                                )
+                            "
```
miggol: On the one hand, I hate that Vue is so implicit. But I must admit that this kind of thing is very easy.

miggol (approved):
As I've already said this is a lovely PR. Really cool to see this working well on the frontend side. 

Plenty to talk about, but the main architectural thing I would like you to take another look at is the role of the `QuestionResponse` to the RSQ.

I can't tell exactly why we need it, hopefully my comments can explain why. If you disagree just give me a call on Teams.

Otherwise, I approve! There's more to do but as this is a PR to a PR I think it's best to keep development speed up rather than to chase perfection. Nice job.

backend/form/models/questions.py
EdoStorm96: I'm down for the custom texts! Was actually already thinking about this. Will get to it!

As far as the `rqq` is concerned: I think we would not need such a model. My vision for this is to just have button next to or below the question itself for delete or add mutation. So the `RepeatableQuestion` is the `RepeatableQuestionQuestion`.

EdoStorm96:
@miggol Yes that is a really good suggestion actually! I've implemented it and also added the custom labels. Have another look at it if you want!

PR #370, opened 2026-08-26
feat: implement dynamic step names

### LINKED ISSUE #292: Question for naming a datasubject group/dynamic stepper bubble name
The PO specified that there needs to be a question to name a datasubject group eg. "Police officer between 20 and 40", "female highschool teachers". Simple textfield. Ideally these names would then get displayed in the stepper, but that will be a lot more involved.

EdoStorm96:
fixes #292 

So here is a way to use user input for stepper names. It was pretty simple to do. A question now has a `step_name_override` Boolean. If true, this question's answer gets used to override the step name. This all happens in the resolver. There can only be on step_name_override per step. I initially tried adding this attribute to `TextQuestion`, but then I could not get the unique constraint to work, so adding it to `BaseQuestion` was a simple solution. I've tested this with repeatable_steps, and the basic idea is fully compatible, so this should be good to go into develop.


backend/form/migrations/0017_basequestion_step_name_override_and_more.py
```diff
@@ -0,0 +1,29 @@
...
+        ("form", "0016_stepinfotext_content_stepinfotext_content_en_and_more"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="basequestion",
+            name="step_name_override",
+            field=models.BooleanField(
+                default=False,
+                help_text="If this is set to True, the answer to the question will override the name of its step (for stepper purposes). It can only be true for one question per step.",
```
XanderVertegaal:
```suggestion
                help_text="If this is set to True, the answer to the question will override the name of its step (for stepper purposes). It can only be True for one question per step.",
```

(Sorry for nitpicking :P)

backend/form/models/questions.py
```diff
@@ -127,6 +139,7 @@ class TrueFalseQuestion(BaseQuestion):
 
 
 class TextQuestion(BaseQuestion):
+
```
XanderVertegaal: Rogue whiteline.

backend/form/services/user_form_resolver.py
```diff
@@ -81,14 +81,25 @@ def _resolve_step_instances(self, step: Step) -> list[StepType]:
...
+            step_name_override = False
             # Get all questions
             questions = []
             for question in step_questions:
-                questions.extend(self._resolve_question_instances(question))
+                resolved_questions = self._resolve_question_instances(question)
+                # We assume that a step_name_override question is never
+                # repeatable, so we work with the first in the list.
+                if question.step_name_override and resolved_questions[0].answer:
+                    # save the question's value in step_name_override var
```
XanderVertegaal: The comment on line 100 can go. (The one on 97f. is an important/good addition!)

backend/form/services/user_form_resolver.py
```diff
@@ -81,14 +81,25 @@ def _resolve_step_instances(self, step: Step) -> list[StepType]:
...
         repeat_count = self.evaluator.get_repeat_count_for_step(step)
         instances = []
 
         for repeat_index in range(repeat_count):
+            step_name_override = False
             # Get all questions
             questions = []
             for question in step_questions:
-                questions.extend(self._resolve_question_instances(question))
+                resolved_questions = self._resolve_question_instances(question)
```
XanderVertegaal: Can `resolved_questions` be empty? In that case `resolved_questions[0]` will raise an IndexError.

backend/form/models/questions.py
```diff
@@ -26,6 +26,13 @@ class BaseQuestion(models.Model):
         validators=[snake_case_validator],
     )
 
+    step_name_override = models.BooleanField(
+        default=False,
+        help_text="If this is set to True, the answer to the question will "
+        "override the name of its step (for stepper purposes). It can only be "
```
XanderVertegaal: Not just for stepper purposes, right? The name entered here is used throughout (also e.g. at the top of the page).

frontend/components/form/TextQuestion.vue
```diff
@@ -27,6 +31,7 @@ const modelValue = defineModel<string>();
             type="text"
             class="form-control"
             :class="{ 'is-invalid': isInvalid }"
+            @blur="emit('blur')"
```
XanderVertegaal: This works well, but I'm a bit conflicted by the fact that we now have two different mechanisms to trigger submission updates whenever questions are updated. Would it be an idea to hook into the existing mechanism with `useWatchQuestions`, using something like this in MRForm.vue?

```ts
// FileUploadQuestions and questions with conditions are watched. If their
// values change, the form is resubmitted.
const watchedQuestions = computed(() =>
    props.step.questions.filter(
        (question) =>
            question.hasConditions ||
            question.__typename === "FileUploadQuestionType" ||
            question.stepNameOverride,
    ),
);
```

Questions with `stepNameOverride` are always(?) TextQuestions, so they should be debounced until the user stops typing for 300ms. This seems to work for me, although we might want to tweak the debounce time (currently: 300 ms). We can then get rid of the blur events and have one central place that watches for 'special' questions that need to trigger a submission.

backend/form/models/questions.py
XanderVertegaal: Doesn't this field belong in `TextQuestion` rather than `BaseQuestion`? I don't see how any other question type could provide a step name. If you have a use case in mind, I'd be happy to hear it.

XanderVertegaal (approved):
This works well. Adding a field and using it in the resolver seems like the correct approach. I'm happy this does not take as many code changes as I feared. Well done! 🚀

I have nothing against merging as is, although I have two things I'd like you to consider: (See my comments below for more details.)
- Move `step_name_override` to `TextQuestion` instead of `BaseQuestion`.
- Use existing `useWatchQuestions` to submit upon question change rather than `blur` events.

(I know I suggested listening to `blur` events, but I now think it may be nicer to use the existing logic as a central funnel, but I'm happy to be convinced otherwise :))

XanderVertegaal:
And an additional request: could you make sure that `create_dev_data` creates a step name override question in some cases, or at least create an issue for it? I read that #371 also does not contain updates to the dev data script, so if you decide to create an issue, you might want to link both this PR and #371.

(I'm a bit worried we'll forget to update create_dev_data and end up with dev data that is not similar to 'real' data anymore.)

backend/form/models/questions.py
EdoStorm96: Yes, I tried to add it to TextQuestion, but then we cannot have the unique constraint, because step is on BaseQuestion. That is why I went this route.

backend/form/models/questions.py
EdoStorm96: Black made me :(

backend/form/models/questions.py
XanderVertegaal: I see... I guess you could also refer to the override question on the Step itself:

```py
class Step(models.Model):
    ...
    step_name_override_question = models.OneToOneField(
        "form.TextQuestion",
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name="override_for_step",
    )
```

That would enforce DB-level uniqueness as well ánd restrict it to TextQuestions, but it would also require refactoring of the logic you've already built. Perhaps it's best to leave it as it is then.

backend/form/models/questions.py
XanderVertegaal: I think Black is okay with it either way. If I remove the whiteline and run Black, it doesn't complain :P Is your Black fussier than mine?

frontend/components/form/TextQuestion.vue
EdoStorm96: I think I agree here. It is a little bit less smooth (for instance if you are keeping a key, or backspace pressed, it will jitter) but I think in real world scenarios it works sufficiently well and it keeps things tidy on our side!

